PR #5, opened 2025-12-28
feat: Add Unicode support for filenames

rawwerks:
## Summary

Previously, the `handelize` function only accepted filenames containing ASCII alphanumeric characters `[a-zA-Z0-9]`. This caused indexing to fail for files with non-ASCII names.

**Before:** Files like `腾讯混元.md`, `日本語.md`, `café-notes.md` would throw an error during indexing.

**After:** Unicode letters and numbers are now supported in filenames across all languages.

## Changes

- Updated validation regex to use Unicode property escapes `[\p{L}\p{N}]` which match any Unicode letter or number
- Updated character replacement regex to preserve Unicode letters/numbers instead of stripping them
- Updated tests to verify Unicode filename support

## Test Plan

- [x] All existing handelize tests pass
- [x] New tests added for Chinese, Japanese, Korean filenames
- [x] Mixed Unicode/ASCII filenames work correctly
- [x] Folder paths with Unicode work correctly

## Example

```typescript
// Before: throws "no valid filename content"
handelize("腾讯混元.md") 

// After: returns "腾讯混元.md"
handelize("腾讯混元.md") // => "腾讯混元.md"
handelize("日本語-notes.md") // => "日本語-notes.md"  
handelize("café-notes.md") // => "café-notes.md"
```

### DIFF
--- a/src/store.test.ts
+++ b/src/store.test.ts
@@ -353,12 +353,16 @@ describe("handelize", () => {
   });
 
   test("handles unicode characters", () => {
-    // Pure unicode with no alphanumerics throws error
-    expect(() => handelize("日本語.md")).toThrow("no valid filename content");
-    // Mixed unicode/ascii preserves the ascii parts
-    expect(handelize("café-notes.md")).toBe("caf-notes.md");
-    expect(handelize("naïve.md")).toBe("na-ve.md");
-    expect(handelize("日本語-notes.md")).toBe("notes.md");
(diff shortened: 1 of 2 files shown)


neoromantic:
@tobi appreciate if you could merge this one!

tobi:
The Unicode filename support was already merged in commit 4d21c5a. Thanks for the contribution - your PR helped inspire the fix!

PR #13, opened 2026-01-09
Fix vector search collection filter

h4rk8s:
- Fixes SQL parameter handling for -c/--collection in vector search
- Prevents no such column: <collection> errors
- Uses proper parameter binding instead of string replacement


tobi:
The collection filter SQL fix was already merged in commit 4d21c5a. Thanks for the bug report and fix!

PR #53, opened 2026-01-27
Fix Nix build with bun2nix integration

jacobmichels:
This PR fixes the Nix build using bun2nix for proper dependency management.

## Problem

The current `flake.nix` attempts to run `bun install` during the build phase, which tries to download packages from the internet and fails.

## Solution

This PR integrates **bun2nix** to create reproducible, pure Nix builds:

- Uses `bun2nix.hook` and `fetchBunDeps` with a fixed output derivation (`bun.nix`)
- Dependencies are locked and fetched declaratively
- Includes nix-community binary cache for pre-built bun2nix binaries
- Sets `dontBuild=true` since qmd runs directly from TypeScript source
- Adds postinstall script to auto-regenerate `bun.nix` when dependencies change

## Testing

```sh
# Build and run
nix build
./result/bin/qmd --help

# Or run directly
nix run github:jacobmichels/qmd/fix/nix-build-with-bun2nix
```

## Result

- ✅ Reproducible, pure builds
- ✅ Dependencies locked in `bun.nix` 
- ✅ Binary cache speeds up builds
- ✅ All functionality preserved
- ✅ Added Nix installation docs to README

PR #55, opened 2026-01-27
Nix fod build

mulatta:
## Summary

Fix Nix build to work in sandboxed environments by using Fixed-Output Derivation (FOD) pattern.

> **Note**: This PR addresses the same issue as #53 but uses a different approach.
> #53 uses `bun2nix` (external flake input), while this PR uses pure nixpkgs FOD pattern without
> additional dependencies.
> Both are valid solutions - maintainer can choose the preferred approach.

### Changes

**Add treefmt-nix formatter**

- Add `treefmt-nix` input for consistent code formatting
- Configure nixfmt for `.nix` files
- Enables `nix fmt` without arguments

**Use FOD for sandboxed bun install**

- Replace network-dependent `bun install` with FOD pattern
- Separate `bunDeps` derivation with platform-specific hashes
- No additional lock files required (uses existing `bun.lock`)

### Problem

The current `flake.nix` runs `bun install` during the build phase, which requires network access and
fails in Nix's sandboxed builds.

### Comparison with #53

| Aspect                | This PR (FOD)             | #53 (bun2nix)               |
| --------------------- | ------------------------- | --------------------------- |
| External flake inputs | treefmt-nix only          | bun2nix                     |
| Generated files       | None                      | `bun.nix`                   |
| Lock file used        | `bun.lock` (existing)     | `bun.lock` → `bun.nix`      |
| Hash management       | Manual (per-platform)     | Auto-generated              |
| Maintenance           | Update hash on dep change | Run `bun2nix` on dep change |

**Trade-offs:**

- **This PR**: Fewer external dependencies, but requires manual hash updates per platform
- **#53**: Automated hash generation via `bun2nix`, but adds external flake input

### Platform Support

| Platform       | Status |
| -------------- | ------ |
| aarch64-darwin | ✅     |
| aarch64-linux  | ✅     |
| x86_64-darwin  | ✅     |
| x86_64-linux   | ✅     |

### Testing

```sh
nix build
./result/bin/qmd --help
nix fmt

Updating Hashes

When dependencies change:
1. Set hash to empty string for your platform
2. Run nix build
3. Copy correct hash from error message

```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -9,10 +9,10 @@
 
   outputs =
     {
-      self,
       nixpkgs,
       flake-utils,
       treefmt-nix,
+      ...
     }:
     flake-utils.lib.eachDefaultSystem (
       system:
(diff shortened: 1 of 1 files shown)


tobi:
this is really neat. can you add a workflow to update the hashes on release tags?

mulatta:
Thanks for the feedback!

Regarding the hash update workflow, I've been thinking about this, and there's a challenge with the current bun FOD approach.

The problem is that Bun FOD requires 4 platform-specific hashes (x86_64-linux, aarch64-linux, x86_64-darwin, aarch64-darwin). thus, to get each hash, you need to run a failing build on that specific platform

However, GitHub Actions has limited darwin runner support (public repos only, or self-hosted). This makes fully automated hash updates impractical

I also submitted this package to [numtide/llm-agents.nix#2134](https://github.com/numtide/llm-agents.nix/pull/2134), and maintainer suggested using `buildNpmPackage` for exactly this reason—single platform-independent `npmDepsHash` enables straightforward CI automation (see: [numtide/llm-agents.nix#2146](https://github.com/numtide/llm-agents.nix/pull/2146))

You may choose one of those options, 
 A. Add npm build option : Single hash, easy workflow, Adds vendored `package-lock.json` 
B. Bun FOD + matrix CI : Pure bun, 4 runners needed, darwin limited
C. Keep manual : Simple, No automation
D. bun2nix : [PR#53](https://github.com/tobi/qmd/pull/53)

I'm happy to implement whichever you prefer. What's your take?

Also: would you consider creating release tags? Both llm-agents.nix and this workflow would benefit from versioned releases rather than tracking HEAD commits.

Mic92:
It's possible to have fod hashes for bun computed with nix in ci, but it basically requires using a build matrix over the 4 architectures to do so and a collector job that than actually updates the hashes file.

PR #57, opened 2026-01-27
fix: prevent disposal crashes and reduce memory usage

### LINKED ISSUE #40: qmd embed fails with DisposedError on slow systems due to inactivity timeout
`qmd embed` fails with `DisposedError: Object is disposed` when embedding is slow (e.g., CPU-only, no GPU acceleration). Most chunks fail, process eventually dies.

**Observed behavior:**
```
Embedding 3500+ documents (33000+ chunks)
██░░░░░░░░░░░░░░░░░░░░░░░░░░░░   7% 19/33165 2573 err
```

~99% failure rate. Error:
```
error: Object is disposed
  at _ensureNotDisposed (node-llama-cpp/.../LlamaContext.js:534:19)
  at dispatchPendingBatch (...)
```

**Root cause:**

`LlamaCpp` has a 2-minute inactivity timeout (llm.ts:229). In `embedBatch()`, `touchActivity()` is called once at the start (llm.ts:545), then `Promise.all` runs all embeddings in parallel. On slow systems where a batch takes longer than 2 minutes, the timer fires mid-batch and disposes the context while embeddings are still pending.

**Fix:**

Call `touchActivity()` after each successful embedding in `embedBatch()`:

```typescript
// llm.ts ~line 550
const embedding = await context.getEmbeddingFor(text);
this.touchActivity();  // Keep-alive during slow batches
return { embedding: Array.from(embedding.vector), model: this.embedModelUri };
```

This correctly signals that work is in progress, preventing premature disposal.

tridip1931:
## Summary

This PR addresses several stability issues affecting systems with limited GPU memory (tested on Apple M1 16GB):

- **Fix DisposedError during slow operations** (Issue #40) - Add keep-alive calls during long-running embed/rerank operations
- **Prevent GC crash during disposal** - Disable auto-disposal to avoid Bun/node-llama-cpp NAPI finalizer bug
- **Reduce memory for query expansion** - Use Qwen3-0.6B (~800MB) instead of 1.7B (~2.2GB)
- **Prevent context overflow in reranking** - Truncate large documents to 1500 chars

## Changes

| File | Change |
|------|--------|
| `src/llm.ts` | Add `touchActivity()` after embeddings and reranking |
| `src/llm.ts` | Disable inactivity timeout for default instance |
| `src/llm.ts` | Switch to smaller Qwen3-0.6B model |
| `src/store.ts` | Truncate documents before reranking (1500 char limit) |

## Test plan

- [x] `qmd query "test" -c collection` completes without crash on M1 16GB
- [x] `qmd embed` should no longer fail with DisposedError on slow systems
- [x] Reranking large documents no longer causes context overflow

## Related Issues

- Fixes #40 (DisposedError on slow systems)

---

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -149,8 +149,8 @@ export type RerankDocument = {
 // Format: hf:<user>/<repo>/<file>
 const DEFAULT_EMBED_MODEL = "hf:ggml-org/embeddinggemma-300M-GGUF/embeddinggemma-300M-Q8_0.gguf";
 const DEFAULT_RERANK_MODEL = "hf:ggml-org/Qwen3-Reranker-0.6B-Q8_0-GGUF/qwen3-reranker-0.6b-q8_0.gguf";
-// const DEFAULT_GENERATE_MODEL = "hf:ggml-org/Qwen3-0.6B-GGUF/Qwen3-0.6B-Q8_0.gguf";
-const DEFAULT_GENERATE_MODEL = "hf:ggml-org/Qwen3-1.7B-GGUF/Qwen3-1.7B-Q8_0.gguf";
+const DEFAULT_GENERATE_MODEL = "hf:ggml-org/Qwen3-0.6B-GGUF/Qwen3-0.6B-Q8_0.gguf";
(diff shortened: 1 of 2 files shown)


tobi:

PR #85, opened 2026-02-01
Harden MCP search path

larusivar:
## Summary\n- Wrap MCP vsearch/query in LLM sessions to prevent idle-dispose errors\n- Add fast/balanced/best mode to MCP query with strong-BM25 fast path\n- Add fallback to BM25-only results on LLM failure\n- Thread LLM session through store expandQuery/rerank/searchVec\n\n## Motivation\nMCP query calls were timing out or failing with "Object is disposed". These changes keep LLM contexts alive during tool calls and avoid heavy expansion/rerank when lexical signals are strong.\n\n## Testing\n- bun -e "import('./src/mcp.ts')"\n

### DIFF
--- a/src/qmd.ts
+++ b/src/qmd.ts
@@ -2143,39 +2143,30 @@ async function querySearch(query: string, opts: OutputOptions, embedModel: strin
     // Map to store hash by filepath for final results
     const hashMap = new Map<string, string>();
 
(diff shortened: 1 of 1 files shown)


larusivar:
Closing per maintainer request. Happy to reopen or revise if needed.

PR #100, opened 2026-02-02
fix: handle non-alphanumeric paths in handelize

chhoumann:
While indexing my notes with:

```sh
qmd collection add ~/notes --name notes
```

I hit a crash that stops the entire indexing run. The immediate trigger was a markdown file named `{.md`.

This is what the failure looked like:

```text
error: handelize: path "{.md" has no valid filename content
    at handelize (.../src/store.ts:787:15)
    at indexFiles (.../src/qmd.ts:1403:18)
    at async collectionAdd (.../src/qmd.ts:1308:9)
```

The underlying cause is that `qmd collection add` normalizes each discovered file path via `handelize(relativeFile)` for token-friendliness, and `handelize` used to reject filenames whose basename without extension contained no Unicode letters or digits. For `{.md`, the basename is `{`, so the check fails and indexing aborts even though the filesystem path is valid.

The fix keeps the existing slugification behavior for normal filenames, but removes the hard rejection for symbol-only segments. Instead, when a path segment would become empty after cleaning, we fall back to a deterministic encoding based on Unicode codepoints. That makes the normalization stable and avoids crashing the indexer on valid filenames.

For example:

```ts
handelize("{.md") // "_u7b.md"
handelize("...")  // "_u2e-2e-2e"
```

I also added tests to cover these edge cases so we do not regress the behavior in the future.

I realize this changes behavior from erroring on some edge-case filenames to always producing a stable, indexable normalized path. I think that tradeoff is worth calling out explicitly because it means certain files that previously could not be indexed will now be indexed under an encoded display path like `_u7b.md`. The encoding is only used when a segment would otherwise normalize to an empty string, and it is deterministic, so the same input path will always map to the same normalized path across runs.

I also increased the timeout on the `chunkDocumentByTokens returns single chunk for small documents` test. On my machine that test occasionally exceeded Bun’s default 5s timeout due to first-use initialization of the tokenizer and native LLM resources, even though the logic under test is fine. The change is just to reduce flakiness, not to mask failures.

Please let me know if there are any questions, feedback, or adjustments! I'd be happy to accommodate them. And please feel free to close if this feels irrelevant.


tobi:

tobi:

PR #106, opened 2026-02-03
Add OpenAI-compatible remote LLM support and remote sanity-check scri…

JasonGiedymin:
Consider this PR as discussion.

The changes were made for the use case where running the text models locally were not possible. i.e. lightweight containers with no pass thru support (you'll find this on Apple with their virtualization and containers framework) or with very limited CPU. By providing a config file, the app will then use a remote server which does have access to a stout CPU or GPU.

* Add remote LLM test script
* Support LM Studio responses rerank, document schema
* Recommend responses-based rerank
* Fix char chunking fallback

Thanks for reading, and thanks to everyone that made this mainline repo a thing. This binary has helped me with some of the PARA framework stuff I'm working on.

### DIFF
--- a/src/qmd.ts
+++ b/src/qmd.ts
@@ -1748,7 +1748,7 @@ type OutputOptions = {
   limit: number;
   minScore: number;
   all?: boolean;
-  collection?: string;  // Filter by collection name (pwd suffix match)
+  collection?: string | string[];  // Filter by collection name(s)
   lineNumbers?: boolean; // Add line numbers to output
   context?: string;      // Optional context for query expansion
 };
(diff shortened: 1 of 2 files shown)


JasonGiedymin:
I've closed this PR #106 as the maintainers have other priorities.

The app has improved performance but for those who still require remote calls (such as myself) my repo will still be available and I'll keep it up to date (within reason). Branch: `remote-api`:  https://github.com/JasonGiedymin/qmd-remote/tree/remote-api

Local perf enhancements in latest mainline:
    QMD now runs on both Node.js and Bun, with up to 2.7x faster reranking
    through parallel GPU contexts. GPU auto-detection replaces the unreliable
    gpu: "auto" with explicit CUDA/Metal/Vulkan probing.

Again, if this isn't possible for you (vps, no GPUs) I'll be maintaining my fork up to date (within reason) and you can continue using that (for now).

PR #117, opened 2026-02-05
fix(vec): avoid pathological join order in sqlite-vec search

dougvk:
Vector search (and the hybrid query pipeline) can become extremely slow on larger indexes because SQLite may choose a join order that scans the documents table before evaluating the vec0 KNN constraint. This makes qmd vsearch/query look like it is hanging even though the embedding call is fast.

This change forces the KNN step to run first by materializing the vec0 lookup in a CTE (WITH knn AS MATERIALIZED ...), then joining the small KNN result set to content_vectors/documents/content. It also removes the placeholder-replacement hack for collection filtering and binds the collection name as a normal parameter.


tobi:

tobi:

PR #116, opened 2026-02-05
feat: Add OpenAI embedding, query expansion, and re-ranking support

jonesj38:
Summary

Optional OpenAI integration for embeddings and query expansion. Dramatically faster for users who prefer API-based inference over local models.

Performance

| Operation                  | Local (llama-cpp) | OpenAI  |
| -------------------------- | ----------------- | ------- |
| Query expansion            | 30-40s            | **200ms**   |
| Full re-embed (30k chunks) | ~2 hours          | **~10 min** |
| Tokenizer load             | 30s               | **0s**      |
| Search latency             | 30-60s            | **3-5s**    |
| Reranking (30 docs)    | 10-15s          | **1-2s**      |`

Features

• OpenAI Embeddings — text-embedding-3-small (1536 dims), native batch API, ~$0.02/1M tokens
• OpenAI Query Expansion — gpt-4o-mini for lex/vec/hyde variants
• OpenAI Reranking — API-based reranking replaces local qwen3-reranker, eliminating model download and GGUF inference overhead
• Tiktoken chunking — eliminates model load time for tokenization
• Robust retry logic — exponential backoff with jitter for rate limits
Usage

`export OPENAI_API_KEY="sk-..."
export QMD_OPENAI=1
qmd embed -f      # Re-embed with OpenAI
qmd search "query"
`

Design

• Opt-in — local models remain the default
• Graceful fallback — errors don't crash, just skip
• Replace local reranking with OpenAI — no GGUF model download or local inference needed
• No breaking changes — existing workflows unchanged
Files Changed

• src/openai-llm.ts — new OpenAI LLM implementation
• src/llm.ts — embedding config, provider switching
• src/store.ts — tiktoken chunking integration
• src/qmd.ts — QMD_OPENAI env var support
Dependencies

• openai — API client
• tiktoken — fast BPE tokenization

PR #65, opened 2026-01-28
feat: add remote embedding providers (Voyage AI, OpenAI-compatible)

adam91holt:
## Summary

Add support for remote embedding APIs as an alternative to local models. This enables:
- Faster embeddings without requiring a GPU
- Higher quality embeddings via Voyage AI
- Flexibility to use any OpenAI-compatible API

### New Features
- **QMD_PROVIDER** env var to select provider: `voyage`, `openai`, or `local` (default)
- **Voyage AI** support: voyage-4-lite embeddings + rerank-2 reranking
- **OpenAI-compatible** support: works with OpenAI, Ollama, vLLM, LM Studio, etc.
- Provider info shown in `qmd status` output

### Technical Details
- New `RemoteLLM` class implementing the `LLM` interface
- `getDefaultLLM()` factory function for provider selection
- Query expansion stays local (LlamaCpp) for all providers
- 15 new tests for remote provider functionality

### Files Changed
- `src/remote.ts` - New RemoteLLM implementation
- `src/remote.test.ts` - Test suite for remote providers  
- `src/llm.ts` - Provider selection functions
- `src/store.ts` - Use getDefaultLLM for embed/rerank
- `src/qmd.ts` - Show provider in status
- `README.md` - Documentation for env vars

## Environment Variables

| Variable | Default | Description |
|----------|---------|-------------|
| `QMD_PROVIDER` | `local` | Provider: `local`, `voyage`, `openai` |
| `VOYAGE_API_KEY` | - | Voyage AI API key |
| `VOYAGE_EMBED_MODEL` | `voyage-4-lite` | Voyage embedding model |
| `VOYAGE_RERANK_MODEL` | `rerank-2` | Voyage reranking model |
| `OPENAI_API_KEY` | - | OpenAI API key |
| `OPENAI_EMBED_MODEL` | `text-embedding-3-small` | OpenAI embedding model |
| `OPENAI_API_BASE` | `https://api.openai.com/v1` | Base URL for OpenAI-compatible APIs |

## Testing

All 15 new tests pass with actual Voyage API calls:
```
bun test src/remote.test.ts
15 pass, 0 fail
```

Tests skip gracefully when API keys are not set.

PR #133, opened 2026-02-08
Add optional OpenRouter provider for embeddings/search pipeline

jamiequint:
## Summary
  - Added provider selection via `QMD_LLM_PROVIDER` with `local` as the default.
  - Added an `OpenRouterLLM` backend for embeddings, query expansion, and reranking.
  - Kept privacy-safe defaults: no remote calls unless explicitly opted in.
  - Added a one-time remote-inference notice per process when OpenRouter mode is active.
  - Added API key loading from:
    - `QMD_OPENROUTER_API_KEY` (preferred)
    - `OPENROUTER_API_KEY`
    - `QMD_OPENROUTER_API_KEY_FILE` (defaults to `~/.config/qmd/openrouter.key`)
  - Made store/CLI LLM calls provider-agnostic.
  - Added char-based chunking fallback for remote providers (no local tokenizer dependency).
  - Added OpenRouter-focused tests and concise README docs for setup and key storage.

## Testing
  ```bash
  ~/.bun/bin/bun test src/llm.openrouter.test.ts
  ~/.bun/bin/bun test src/llm.test.ts --test-name-pattern "Default LlamaCpp Singleton|LlamaCpp.modelExists"
  ~/.bun/bin/bun src/qmd.ts --help

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 An on-device search engine for everything you need to remember. Index your markdown notes, meeting transcripts, documentation, and knowledge bases. Search with keywords or natural language. Ideal for your agentic flows.
 
-QMD combines BM25 full-text search, vector semantic search, and LLM re-ranking—all running locally via node-llama-cpp with GGUF models.
+QMD combines BM25 full-text search, vector semantic search, and LLM re-ranking. By default it runs locally via node-llama-cpp with GGUF models, with optional OpenRouter-backed inference.
 
 ![QMD Architecture](assets/qmd-architecture.png)
 
@@ -280,6 +280,36 @@ Supported model families:
 > since vectors are not cross-compatible between models. The prompt format is
 > automatically adjusted for each model family.
 
+### OpenRouter Mode (Optional)
+
+QMD defaults to `local` inference. To use OpenRouter for embeddings, query expansion, and reranking:
+
+```sh
+export QMD_LLM_PROVIDER=openrouter
+export QMD_OPENROUTER_API_KEY="sk-or-..."
+```
+
+When OpenRouter mode is active, QMD prints a single remote-inference notice per process.
+
+You can also store the key in a file (default path):
+
+```sh
+mkdir -p ~/.config/qmd
+chmod 700 ~/.config/qmd
+printf '%s\n' 'sk-or-...' > ~/.config/qmd/openrouter.key
+chmod 600 ~/.config/qmd/openrouter.key
+export QMD_LLM_PROVIDER=openrouter
+```
+
+Supported key env vars:
+- `QMD_OPENROUTER_API_KEY` (preferred)
+- `OPENROUTER_API_KEY`
+- `QMD_OPENROUTER_API_KEY_FILE` (custom key file path)
+
+Optional model overrides:
+- `QMD_OPENROUTER_EMBED_MODEL`
+- `QMD_OPENROUTER_GENERATE_MODEL`
+- `QMD_OPENROUTER_RERANK_MODEL`
 ## Installation
 
 ```sh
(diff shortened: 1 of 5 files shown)


alexleach:
I feel like making this specific for OpenRouter reduces the usefulness of this PR. OpenRouter is itself OpenAI-compatible, so it would make better sense to map this PR around the OpenAI-spec, not OpenRouter's implementation of it. A rename of accepted environment variables is all that is required, really:-

```
DEFAULT_OPENROUTER_BASE_URL ⟶ QMD_OPENAI_BASE_URL
QMD_OPENROUTER_API_KEY ⟶ QMD_OPENAI_API_KEY
QMD_OPENROUTER_API_KEY_FILE ⟶ QMD_OPENAI_API_KEY_FILE
```
For `QMD_LLM_PROVIDER`, a selector between `local` and `remote` or `openai` would be more generic.

I see you have other environment variables, too:-

```
QMD_OPENROUTER_EMBED_MODEL
QMD_OPENROUTER_GENERATE_MODEL 
QMD_OPENROUTER_RERANK_MODEL 
QMD_OPENROUTER_APP_NAME
QMD_OPENROUTER_APP_URL
```

The first three would easily map to `QMD_OPENAI_*_MODEL` variables.

I'm not sure the purpose of the `*_APP_NAME` or `_APP_URL` variables, though. They look a bit unnecessary, at first glance?

jamiequint:
@alexleach Thanks for the feedback. These are OpenRouter-specific metadata fields and aren’t required for inference, so removing them seems reasonable:

QMD_OPENROUTER_APP_NAME
QMD_OPENROUTER_APP_URL

I went ahead and did that.

On the broader point about generalizing this away from OpenRouter to OpenAI: there isn’t anything in the code that requires OpenAI embeddings specifically. This implementation allows users to choose non-OpenAI embedding models via `QMD_OPENROUTER_EMBED_MODEL` if they want.

Because of that, names like `QMD_OPENAI_BASE_URL`, `QMD_OPENAI_API_KEY`, and `QMD_OPENAI_API_KEY_FILE` would imply an OpenAI-specific coupling that does not actually exist here.

alexleach:
Thanks for the response. What I meant was the protocol itself is basically OpenAI's. From https://openrouter.ai/docs/api/reference/overview:-

> OpenRouter’s request and response schemas are very similar to the OpenAI Chat API, with a few small differences. At a high level, OpenRouter normalizes the schema across models and providers so you only need to learn one.

So, while not identical, they are "very similar", and one PR could satisfy both OpenAI and OpenRouter. Being OpenAI who designed the API, I would have thought naming it that way would be more applicable.

Personally, I run qmd's default models in [Docker Model Runner](https://www.docker.com/blog/docker-model-runner-vllm-metal-macos/), so that openclaw running in docker can use qmd with GPU acceleration on my Mac. Docker Model Runner also conforms to the OpenAI Chat API. So, I myself use neither OpenAI or OpenRouter, but your PR would presumably cover my use-case too.

alexleach:
There are at least a couple of other implementations of this as well:-

- https://github.com/tobi/qmd/pull/65
- https://github.com/tobi/qmd/pull/116

I'm hoping one of them gets merged so I don't have to rely on a fork of a fork, which is what I'm doing now...

gregorycollins-usmobile:
I needed a patch to get this branch working https://github.com/tobi/qmd/commit/5964419d7f7e9409fa9e705023cba4f99c154637, but it works --- second one I tried to get going

gregorycollins-usmobile:
I also had to [kludge in a better rerank](https://github.com/gregorycollins-usmobile/qmd/commits/feat/openrouter-provider/) because OpenRouter doesn't currently support it

tobi:

tobi:

PR #137, opened 2026-02-08
fix: Windows compatibility for HOME path, mkdir, and sqlite-vec loading

sarukas:
## Summary

- **HOME path fallback**: Use `USERPROFILE` env var when `HOME` is not set (Windows cmd.exe doesn't set `HOME`, only git-bash does), preventing the database from being created at `/tmp/.cache/qmd/` instead of the user's home directory
- **Cross-platform mkdir**: Replace Unix-only `Bun.spawnSync(["mkdir", "-p", ...])` with Node.js `mkdirSync(..., { recursive: true })` which works on all platforms
- **sqlite-vec DLL loading**: Add Windows-specific fallback to load `vec0.dll` via absolute path from `node_modules/sqlite-vec-windows-x64/` when the default loader fails (happens when running from outside the project directory via `bun link`)
- **bin entry**: Point to `./src/qmd.ts` instead of `./qmd` shell script so `bun link` creates a working Windows shim

## Test plan

- [x] `qmd search` works from any directory on Windows cmd.exe
- [x] `qmd vsearch` works from any directory on Windows cmd.exe
- [x] `qmd collection list` works from any directory on Windows cmd.exe
- [x] All changes are guarded to only affect Windows (non-Windows paths unchanged)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "Quick Markdown Search - Full-text and vector search for markdown files",
   "type": "module",
   "bin": {
-    "qmd": "./qmd"
+    "qmd": "./src/qmd.ts"
   },
   "scripts": {
     "test": "bun test",
--- a/src/store.ts
+++ b/src/store.ts
@@ -13,7 +13,8 @@
 
 import { Database } from "bun:sqlite";
 import { Glob } from "bun";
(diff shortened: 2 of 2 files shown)


tobi:

tobi:

PR #148, opened 2026-02-09
fix(mcp): chunk documents before reranking in query tool

kephail:
## Problem

The MCP `query` tool passes full document bodies to the reranker, which crashes on large documents (e.g. session transcripts >500KB) with:

```
The input lengths of some of the given documents exceed the context size.
Try to increase the context size to at least 509038
```

This also causes Bun to segfault under memory pressure (peak RSS ~7.8GB on a 16GB machine).

## Root Cause

The CLI `querySearch` already handles this correctly — it calls `chunkDocument()` on each candidate and picks the best chunk per document via keyword overlap before passing to the reranker (lines 2200-2220 in `qmd.ts`).

The MCP `query` tool was missing this step, passing `c.body` directly to `store.rerank()`.

## Fix

Apply the same chunking strategy to the MCP query tool:
1. Chunk each candidate document via `chunkDocument()`
2. Select the best chunk per document using keyword overlap with the query
3. Pass chunks (not full bodies) to the reranker

This matches the existing CLI behaviour and keeps documents within the reranker's context window.

## Testing

Reproduced with OpenClaw's QMD session transcript indexing — a 1.6MB session transcript was causing the crash. After this fix, the MCP server stays stable and reranking works correctly on large documents.

### DIFF
--- a/src/mcp.ts
+++ b/src/mcp.ts
@@ -15,6 +15,7 @@ import {
   createStore,
   reciprocalRankFusion,
   extractSnippet,
+  chunkDocument,
   DEFAULT_EMBED_MODEL,
   DEFAULT_QUERY_MODEL,
   DEFAULT_RERANK_MODEL,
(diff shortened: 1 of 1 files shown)


tobi:
This was already fixed when the chunking logic was moved into hybridQuery() in store.ts — the MCP deep_search tool now delegates to hybridQuery which chunks before reranking. No separate fix needed.

PR #179, opened 2026-02-15
Fix tsx runtime dependency resolution

0xble:
## Problem

The qmd wrapper script uses `--import tsx` with a bare specifier, which Node.js resolves relative to the current working directory instead of the package installation directory. This causes `ERR_MODULE_NOT_FOUND` errors when tsx is not installed globally.

Additionally, tsx is currently listed as a devDependency, but it's required at runtime by the wrapper script.

## Solution

This PR implements the fix recommended by analyzing the issue:

1. **Move tsx to runtime dependencies** - tsx is required at runtime, not just development
2. **Fix wrapper script tsx resolution** - Added `find_tsx_loader()` function that searches up from the package installation directory to find `node_modules/tsx/dist/loader.mjs`, working with both nested and hoisted node_modules layouts
3. **Update documentation** - Clarified in CLAUDE.md that:
   - Development can use Bun for package management and testing
   - Runtime requires Node.js >=22 due to better-sqlite3 incompatibility with Bun
   - The wrapper uses Node.js + tsx, not Bun

## Changes

- **package.json**: Moved `tsx` from `devDependencies` to `dependencies`
- **qmd wrapper**: Added robust tsx loader resolution that works regardless of installation method
- **CLAUDE.md**: Updated to clarify Node.js runtime requirement vs Bun development usage

## Testing

After these changes, `qmd` will work correctly when installed via:
- `npm install -g @tobilu/qmd`
- `bun install -g @tobilu/qmd`

Without requiring a separate global tsx installation.

Fixes the ERR_MODULE_NOT_FOUND error reported by users installing qmd.

### DIFF
--- a/src/qmd.ts
+++ b/src/qmd.ts
@@ -60,6 +60,7 @@ import {
   vectorSearchQuery,
   addLineNumbers,
   type ExpandedQuery,
+  type SearchResult,
   DEFAULT_EMBED_MODEL,
   DEFAULT_RERANK_MODEL,
   DEFAULT_GLOB,
@@ -1728,6 +1729,8 @@ function normalizeBM25(score: number): number {
   return 1 / (1 + Math.exp(-(absScore - 5) / 3));
 }
 
+type FieldsLevel = 'minimal' | 'summary' | 'full';
+
 type OutputOptions = {
   format: OutputFormat;
   full: boolean;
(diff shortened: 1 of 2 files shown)


JannikWempe:
`qmd` is currently broken for me because of the mentioned issue.

PR #251, opened 2026-02-23
fix(mcp): support multiple concurrent client sessions

Kaylebor:
Store transports in Map by session ID and create new transport per initialize request. This allows multiple MCP clients to connect simultaneously to the same QMD server.

Also fix README to document actual tool names (4 tools, not 6).


Kaylebor:
I fixed it @kleddbot

Kaylebor:
Just to clarify - it's closed because it's being addressed in other PR right? 👀

PR #258, opened 2026-02-24
fix(llm): try all available GPU types before falling back to CPU

arthurbragaa:
## Problem

On systems where `getLlamaGpuTypes()` reports multiple GPU types but the highest-priority one fails to initialize, `ensureLlama()` skips all remaining GPU types and falls directly to CPU.

**Concrete scenario:** AMD Radeon 890M with Vulkan support, no CUDA toolkit installed.

- `getLlamaGpuTypes()` returns `["cuda", "vulkan"]` (cuda is reported because the prebuilt binary metadata advertises it)
- `ensureLlama()` picks `"cuda"` (first match from `find()`)
- `getLlama({ gpu: "cuda" })` fails — triggers a full CMake build attempt for CUDA, which fails because no CUDA toolkit is present
- Falls directly to `getLlama({ gpu: false })` — CPU mode
- **Vulkan is never tried**, even though it works perfectly on this system

## Fix

Change the GPU selection from "try first match, then CPU" to "try each available GPU in priority order, then CPU":

- `find()` → `filter()` to get **all** available GPU types
- Iterate with a `for` loop, `break` on first success
- Only fall to CPU when the loop exhausts all options
- Each failed GPU type emits a warning (existing behavior preserved)

Priority order is unchanged: `cuda > metal > vulkan > CPU`.

## Testing

Verified on AMD Radeon 890M (RDNA 4, 16GB VRAM) with `vulkan-radeon` driver:

- **Before:** cuda reported → CMake CUDA build fails → CPU fallback (slow, ~minutes for embeddings)
- **After:** cuda reported → fails → vulkan tried → succeeds → GPU acceleration with full VRAM offloading

```
Device
  GPU:      vulkan (offloading: yes)
  Devices:  AMD Radeon 890M Graphics (RADV GFX1150)
  VRAM:     16.2 GB free / 17.5 GB total
```

Embedding 98 chunks completes in 7s on Vulkan vs minutes on CPU.

This is a one-line semantic change (`find` → `filter` + loop), preserving all existing behavior for systems where the first GPU type succeeds.

### DIFF
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -497,25 +497,33 @@ export class LlamaCpp implements LLM {
    */
   private async ensureLlama(): Promise<Llama> {
     if (!this.llama) {
-      // Detect available GPU types and use the best one.
-      // We can't rely on gpu:"auto" — it returns false even when CUDA is available
+      // Detect available GPU types and try each in preference order.
+      //
+      // We can't rely on gpu:"auto" — it returns false even when GPUs are available
       // (likely a binary/build config issue in node-llama-cpp).
+      //
+      // getLlamaGpuTypes() can report GPU types that fail to initialize at runtime
+      // (e.g. "cuda" is reported on systems with no CUDA toolkit installed, because
+      // the prebuilt binary metadata advertises it). We iterate all available types
+      // so that e.g. a Vulkan-capable system isn't forced to CPU just because CUDA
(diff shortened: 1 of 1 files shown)


arthurbragaa:
There are literally 10s of PRs solving this problem already. I'll close mine

PR #284, opened 2026-03-03
fix(http): support multiple concurrent MCP sessions

nnorris7:
## Summary

- The HTTP transport creates a single `McpServer` + `WebStandardStreamableHTTPServerTransport` pair and connects them at startup. Once one client initializes a session, any subsequent client gets `"Server already initialized"` (HTTP 400) because the transport's `_initialized` flag is already set.
- This breaks multi-client setups — e.g. Claude Code holds a persistent session to QMD while a dashboard also needs to search.
- Fix: create a per-session server+transport pair using the SDK's `onsessioninitialized`/`onsessionclosed` callbacks. Each `initialize` request spawns a fresh pair; subsequent requests route to the correct session via the `mcp-session-id` header. All sessions share the same underlying store (read-only search, no conflicts).

## Test plan

- [x] Two concurrent `initialize` requests return distinct session IDs (previously second would 400)
- [x] Tool calls route correctly by session ID
- [x] Existing `vitest` MCP HTTP transport tests pass (56/56)
- [x] Shutdown closes all active sessions

### DIFF
--- a/src/mcp.ts
+++ b/src/mcp.ts
@@ -552,12 +552,30 @@ export type HttpServerHandle = {
  */
 export async function startMcpHttpServer(port: number, options?: { quiet?: boolean }): Promise<HttpServerHandle> {
   const store = createStore();
-  const mcpServer = createMcpServer(store);
-  const transport = new WebStandardStreamableHTTPServerTransport({
-    sessionIdGenerator: () => randomUUID(),
-    enableJsonResponse: true,
-  });
-  await mcpServer.connect(transport);
+
+  // Per-session transport+server pairs. Each MCP client gets its own isolated
+  // session so multiple clients (e.g. Claude Code + dashboard) can connect
+  // concurrently without "Server already initialized" errors.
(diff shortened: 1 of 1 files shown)


nnorris7:
Note: I noticed #251 addresses the same single-session limitation. Our approach differs in a few ways based on the issues identified in that PR's review:

- **Per-session `McpServer` instances** — we call `createMcpServer(store)` for each session rather than reusing a single server. The SDK's `Protocol.connect()` assumes single-transport ownership, so sharing one server across transports causes response misrouting (issue #2 in #251's review).
- **Separate POST vs GET/DELETE handlers** — we preserved the original two-block structure for `/mcp` routing, which avoids the body-parse crash on non-POST methods (issue #1 in #251's review).
- **Shared store** — all sessions share the same read-only store, so there's no resource duplication beyond the lightweight server+transport wrapper per session.

One gap we share with #251: neither implementation has a TTL or max-sessions cap for orphaned sessions (clients that disconnect without a clean close). For a localhost-only tool this is low risk, but happy to add a periodic sweep if desired.

Issue #87, opened 2026-02-01
Emebeding fails on NixOS (node-llama-cpp)


jeffutter:
I found some other issues suggesting the flake won't install. I can't say that I had that experience. I got `qmd` to install just fine from the flake. However when I go to embed my collections I get the following error:

```
❯ qmd embed
Chunking 1243 documents by token count...
[node-llama-cpp] The prebuilt binaries cannot be used in this Linux distro, as `glibc` is not detected
[node-llama-cpp] Failed to build llama.cpp with no GPU support. Error: EROFS: read-only file system, mkdir '/nix/store/yg6xz4rk63c8mksmp9444lyxy6ncxdj1-qmd-1.0.0/lib/qmd/node_modules/node-llama-cpp/llama/localBuilds'
    path: "/nix/store/yg6xz4rk63c8mksmp9444lyxy6ncxdj1-qmd-1.0.0/lib/qmd/node_modules/node-llama-cpp/llama/localBuilds",
 syscall: "mkdir",
   errno: -30,
    code: "EROFS"


EROFS: read-only file system, mkdir '/nix/store/yg6xz4rk63c8mksmp9444lyxy6ncxdj1-qmd-1.0.0/lib/qmd/node_modules/node-llama-cpp/llama/localBuilds'
    path: "/nix/store/yg6xz4rk63c8mksmp9444lyxy6ncxdj1-qmd-1.0.0/lib/qmd/node_modules/node-llama-cpp/llama/localBuilds",
 syscall: "mkdir",
   errno: -30,
    code: "EROFS"


Bun v1.3.6 (Linux x64)
```

I think there may be two issues here.1. It can't detect `glib` on nixos. 2. I'm not sure if it would be able to compile llama.cpp due to the nix store being read-only.

I haven't had a chance to figure out a good path to fix this, but I did find a couple flakes other folks have written that seem to avoid it:
- https://github.com/numtide/llm-agents.nix/blob/59146352b94f16ac823bd1d17f23d78c54ae8952/packages/qmd/package.nix#L147
- https://github.com/MichaelVessia/nixos-config/blob/adcef36314ffea31f0f47a301a354cf7ecf4ad84/modules/programs/qmd/default.nix#L24

PR #285, opened 2026-03-04
fix(nix): add flake-based Nix build with bun2nix

### LINKED ISSUE #223: Nix flake build fails in sandboxed environments (network access during buildPhase)
The current flake.nix fails to build in standard Nix sandboxed environments because bun install is called during the buildPhase, which attempts to download dependencies from the network.

  Error
```
  qmd> Running phase: buildPhase
  qmd> bun install v1.3.9 (cf6cdbbb)
  qmd> error: ConnectionRefused downloading tarball typescript@5.9.3
  qmd> error: ConnectionRefused downloading tarball node-llama-cpp@3.14.5
  ... (hundreds of similar errors)
```

  Nix builds run in a sandbox without network access by default. The current buildPhase attempts to fetch packages at build time:

```
  buildPhase = ''
    export HOME=$(mktemp -d)
    bun install --frozen-lockfile
  '';
```

  Environment

  - NixOS 26.05
  - nix.settings.sandbox = true (default)
  - Using qmd as a flake input in another flake

Cody-W-Tucker:
## Summary

- Add flake.nix with bun2nix overlay for reproducible builds
- Add nix/bun.nix with prefetched npm dependencies  
- Add nix/README.md with installation and maintenance docs
- Add GitHub Action to auto-update bun.nix on bun.lock changes

## Problem

Since Nix flakes are built in sandboxed environments where network access is restricted during the buildPhase. The old flake tries to download npm packages at build time, which fails because the network is unavailable.

## Solution

Use bun2nix to convert Bun dependencies into Nix derivations, prefetching all npm packages into the Nix store at eval time. This allows the build to work completely offline.

Fixes #223


Cody-W-Tucker:
I'm closing this because [numtide/llm-agents.nix](https://github.com/numtide/llm-agents.nix) provides better Nix support for QMD and CUDA works there.

Nix users should consider using installing QMD via [their package](https://github.com/numtide/llm-agents.nix/tree/main/packages/qmd). 

Alternatively, we could emulate their approach here. But I'll leave that for someone else.

Related issues: #87, #223

Issue #222, opened 2026-02-19
Windows ARM64 + Node 25: qmd embed fails due to node-llama-cpp CUDA build + sqlite-vec unavailable


MrHinsh:
## Summary

On Windows ARM64, qmd is effectively unusable out of the box.

Running `qmd embed` triggers:
1. A failing CUDA build attempt via node-llama-cpp (LLVM for WoA required).
2. A hard failure due to `sqlite-vec` not being available.

This makes embedding impossible on a clean Windows ARM64 system.

## Environment

- OS: Windows 11 ARM64 (WoA)
- Architecture: win32 arm64
- Node: v25.5.0
- npm global install: @tobilu/qmd@1.0.7
- GPU: none (CPU-only environment)

`node -p "process.platform + ' ' + process.arch"`
→ win32 arm64

## Reproduction

1. npm install -g @tobilu/qmd
2. qmd embed

## Observed Behaviour

First, node-llama-cpp attempts to build a CUDA variant:

[node-llama-cpp] A prebuilt binary was not found, falling back to building from source
CMake Error:
  LLVM installation was not found. Please install LLVM for WoA

Then, even after falling back to CPU:

Error: sqlite-vec is not available.
Vector operations require a SQLite build with extension loading support.

## Problems

1. The CLI attempts a CUDA build on Windows ARM64 even when no GPU is present.
2. No prebuilt binaries appear to be available for win-arm64 + current Node ABI.
3. sqlite-vec is not available on this platform, causing embed to fail completely.
4. There is no documented Windows ARM64 setup path.

## Expected Behaviour

- CPU-only mode should work without requiring LLVM or CUDA toolchains.
- sqlite-vec dependency should either:
  - have a prebuilt for win-arm64, or
  - degrade gracefully with a clear instruction path.

## Suggestions

- Publish win-arm64 prebuilds for:
  - node-llama-cpp dependency stack
  - sqlite-vec dependency
- Do not attempt CUDA builds when no GPU backend is configured.
- Document supported Node versions and platforms explicitly.

At present, Windows ARM64 appears unsupported in practice.

tobi:
Closing as part of the post-v2.5.1 backlog cleanup: this item is over two months old and has either been superseded by newer QMD releases or is too stale to keep actionable. If it still reproduces on v2.5.1 or newer, please open a fresh focused issue with current repro steps.

MrHinsh:
> has either been superseded by newer QMD releases or is too stale to keep actionable

[...]

PR #310, opened 2026-03-06
feat: use `build: "autoAttempt"` on `getLlama`

### LINKED ISSUE #59: SIGABRT crash on exit during Metal cleanup (macOS ARM64)
## Environment
- macOS ARM64 (Apple Silicon)
- bun 1.1.45
- node-llama-cpp 3.15.1

## Issue
`qmd query` crashes with SIGABRT during process exit after successfully completing reranking:

```
GGML_ASSERT([rsets->data count] == 0) failed
```

**The reranker works correctly** — results are output before the crash. The crash happens during Metal device cleanup in llama.cpp's `ggml_metal_device_free()`.

## Reproduction
```bash
qmd query "winback" -n 3
# Results print successfully, then crash on exit
```

## Root cause
This is an upstream llama.cpp bug in Metal cleanup. Reference: https://github.com/ggml-org/llama.cpp/pull/17869

## Workaround
Redirect stderr and force exit 0:
```bash
qmd query "$@" 2>/dev/null; exit 0
```

## Impact
- CLI usage: cosmetic (results complete before crash)
- MCP server: may affect long-running server stability

Would be great to either:
1. Update to a llama.cpp version with the fix
2. Add explicit resource cleanup before exit to avoid the race

### LINKED ISSUE #87: Emebeding fails on NixOS (node-llama-cpp)

### LINKED ISSUE #185: macOS Intel: unnecessary CUDA build attempt causes ~30s delay on every startup
## Environment

- **OS**: macOS (Intel x86_64, i9-8950HK)
- **Bun**: v1.3.9
- **qmd**: @tobilu/qmd@1.0.0
- **node-llama-cpp**: 3.15.1

## Problem

On every `qmd query` invocation, `node-llama-cpp` attempts to build a CUDA binary from source, which always fails on macOS (no NVIDIA GPU), adding ~30 seconds of noisy error output before falling back to CPU.

Output on every startup:

```
[node-llama-cpp] A prebuilt binary was not found, falling back to building from source
...
CMake Error at llama.cpp/ggml/src/ggml-cuda/CMakeLists.txt:258 (message):
  CUDA Toolkit not found
...
[node-llama-cpp] Failed to build llama.cpp with CUDA support.
QMD Warning: cuda reported available but failed to initialize. Falling back to CPU.
QMD Warning: no GPU acceleration, running on CPU (slow). Run 'qmd status' for details.
```

## Root Cause

In `llm.ts`, `ensureLlama()` calls `getLlamaGpuTypes()` which incorrectly reports CUDA as available on macOS Intel. The code then tries `getLlama({ gpu: "cuda" })`, triggering a full CMake build from source that always fails:

```ts
const gpuTypes = await getLlamaGpuTypes();
const preferred = (["cuda", "metal", "vulkan"] as const).find(g => gpuTypes.includes(g));
```

**Facts about macOS Intel (x86_64)**:
- ❌ CUDA — never available (NVIDIA dropped Mac support years ago)
- ❌ Metal GPU acceleration — only on Apple Silicon
- ✅ CPU prebuilt binary — already exists at `@node-llama-cpp/mac-x64`

## Suggested Fix

Skip GPU detection entirely on macOS x86_64:

```ts
let preferred: "cuda" | "metal" | "vulkan" | undefined;
if (process.platform === "darwin" && process.arch === "x64") {
  preferred = undefined; // Intel Mac: always CPU
} else {
  const gpuTypes = await getLlamaGpuTypes();
  preferred = (["cuda", "metal", "vulkan"] as const).find(g => gpuTypes.includes(g));
}
```

## Impact

- Removes ~30s startup delay on every invocation
- Eliminates ~50 lines of noisy CMake error output
- CPU prebuilt binary is already cached and works correctly

### LINKED ISSUE #194: Performance: Vulkan library detection adds 2-3s to every invocation on machines without a GPU
## Problem

On a Linux x86_64 machine without a GPU, every `qmd` command that uses llama.cpp (vsearch, query, embed, status) spends 2-3 seconds trying to build/load with Vulkan support before falling back to CPU.

### Root cause

`node-llama-cpp` detects `vulkan: true` because `libvulkan.so.1` exists on the system (installed as a Mesa/graphics stack dependency). There is no actual GPU, but the library presence is enough to trigger the Vulkan build path.

The qmd code in `ensureLlama()` then picks Vulkan as the preferred GPU type:

```js
const preferred = ["cuda", "metal", "vulkan"].find(g => gpuTypes.includes(g));
```

This triggers a cmake build attempt with Vulkan, which either fails or produces a binary that cannot initialize — then falls back to `getLlama({ gpu: false })`. This happens on **every invocation** since the failure is not cached.

On a machine where vsearch takes ~1.4s of actual work, the probe adds ~2s of overhead (3.5s total, 1.4s after patching).

## Reproduction

```bash
# On a Linux machine with no GPU but libvulkan.so installed (common on Ubuntu/Debian)
time qmd vsearch "test query" -n 1
# Observe cmake/build output in stderr before results
```

## Suggested fix

A few options:
1. **Environment variable** like `QMD_GPU=false` or `QMD_NO_GPU=1` to skip the probe entirely
2. **Cache the probe result** — if Vulkan init fails once, remember that and skip on subsequent runs
3. **Runtime device check** — having Vulkan libraries installed ≠ having a usable Vulkan device. Check for actual enumerable devices before attempting the build

## Workaround

Patching `dist/llm.js` to replace the GPU detection block with:
```js
let llama = await getLlama({ gpu: false, logLevel: LlamaLogLevel.error });
```

## Environment
- qmd 1.0.6
- node-llama-cpp (bundled)
- Linux x86_64, no GPU, libvulkan.so.1 present (Mesa), 28GB RAM
- Node v22

### LINKED ISSUE #213: GPU fallback skips Vulkan when CUDA fails — falls directly to CPU
## Bug Description

On Linux x64 systems with Vulkan GPU (Intel Arc) but no CUDA, qmd falls back to CPU instead of trying Vulkan. The GPU fallback chain is `CUDA → CPU` instead of `CUDA → Vulkan → CPU`.

## Root Cause

In `src/llm.ts`, the `ensureLlama()` method calls `getLlamaGpuTypes()` **without** the `"supported"` argument. Without `"supported"`, `getLlamaGpuTypes()` returns all valid GPU types for the platform (`["cuda", "vulkan", false]` on Linux x64) — **regardless of actual hardware**.

qmd then picks the first match (`"cuda"` — highest priority), calls `getLlama({gpu: "cuda"})`, it fails (no CUDA toolkit), and the `catch` block falls directly to `getLlama({gpu: false})` (CPU) — **skipping Vulkan entirely**.

```javascript
// Current code (llm.ts ~line 248-262)
const preferred = ["cuda", "metal", "vulkan"].find(g => gpuTypes.includes(g));
if (preferred) {
    try {
        llama = await getLlama({ gpu: preferred, logLevel: LlamaLogLevel.error });
    } catch {
        // BUG: jumps straight to CPU, never tries vulkan
        llama = await getLlama({ gpu: false, logLevel: LlamaLogLevel.error });
    }
}
```

## Fix

Replace `find()` (picks one) with `filter()` (keeps all) and loop through GPU types:

```javascript
const gpuPriority = ["cuda", "metal", "vulkan"].filter(g => gpuTypes.includes(g));
let llama;
let initialized = false;
for (const gpuType of gpuPriority) {
    try {
        llama = await getLlama({ gpu: gpuType, logLevel: LlamaLogLevel.error });
        initialized = true;
        break;
    } catch {
        process.stderr.write(`QMD Warning: ${gpuType} failed. Trying next.\n`);
    }
}
if (!initialized) {
    llama = await getLlama({ gpu: false, logLevel: LlamaLogLevel.error });
}
```

## Environment

- qmd 1.0.6
- node-llama-cpp 3.15.1
- Linux x64 (Ubuntu 24.04, kernel 6.17)
- Intel Core Ultra 7 255H with Intel Arc (Arrow Lake-P) integrated GPU
- Vulkan 1.4.318 (Mesa 25.2.8) — working (`vulkaninfo` confirms device)
- No NVIDIA GPU, no CUDA toolkit

## Verification

After patching `dist/llm.js` locally, qmd correctly falls through to Vulkan:

```
QMD Warning: cuda reported available but failed to initialize. Trying next GPU type.

Device
  GPU:      vulkan (offloading: yes)
  Devices:  Intel(R) Graphics (ARL)
  VRAM:     20.4 GB free / 22.7 GB total
```

Before the patch: `GPU: none (running on CPU — models will be slow)` with ~7h ETA for 21K chunks.

## Additional Note

A secondary issue: `getLlamaGpuTypes()` without `"supported"` returns platform-valid types, not hardware-available types. Calling `getLlamaGpuTypes("supported")` would use `detectAvailableComputeLayers()` which correctly returns `{cuda: {hasNvidiaDriver: false, hasCudaRuntime: false}, vulkan: true}`. Using `"supported"` would avoid the CUDA attempt entirely on non-NVIDIA systems. However, the comment in the source suggests this was intentional due to a node-llama-cpp issue with `gpu:"auto"`.

### LINKED ISSUE #268: getLlamaGpuTypes() falsely detects CUDA/Vulkan on macOS x64, causing build-from-source fallback
## Problem

On macOS x64 (Intel), `qmd` falls back to building `node-llama-cpp` from source on every invocation:

```
[node-llama-cpp] A prebuilt binary was not found, falling back to building from source
```

## Root Cause

In `llm.ts`'s `ensureLlama()`, `getLlamaGpuTypes()` incorrectly reports `["cuda", "vulkan", false]` on macOS x64 (Intel i5-1038NG7). The code then picks `"cuda"` as preferred and calls `getLlama({ gpu: "cuda" })`. Since there's no `@node-llama-cpp/mac-x64-cuda` prebuilt package, `node-llama-cpp` attempts to build from source before the `catch` block falls back to CPU.

The CPU prebuilt at `@node-llama-cpp/mac-x64` works perfectly — it's just never reached on the happy path.

## Environment

- macOS (Darwin 25.3.0), Intel x86_64
- qmd 1.0.7
- node-llama-cpp 3.16.2
- Node.js v25.6.1

## Suggested Fix

Skip GPU detection on `darwin` + `x64` since there's no GPU-accelerated prebuilt for that platform anyway. Or support a `QMD_GPU` env var override:

```ts
const isMacX64 = process.platform === "darwin" && process.arch === "x64";
if (isMacX64) {
    llama = await getLlama({ gpu: false, logLevel: LlamaLogLevel.error });
} else {
    // existing GPU detection logic
}
```

## Workaround

Patching `dist/llm.js` to force `gpu: false` on mac x64 resolves the issue — the CPU prebuilt loads instantly without any build step.

### LINKED ISSUE #287: GPU detection issue - goes to CPU even if vulkan is available
I'm on an AMD GPU (RX 7900 XTX) with Vulkan available. `node-llama-cpp inspect gpu` sees it fine, but qmd always ends up on CPU.

It seems the main issue is that `getLlamaGpuTypes()` is called without an argument.

In `llm.js`, `getLlamaGpuTypes()` is called with no argument. When `include` is undefined, node-llama-cpp doesn't actually check what hardware is present, it just hardcodes `["cuda", "vulkan", false]` on any Linux system (see `getLlamaGpuTypes.js` lines 24-32). The fix is to call `getLlamaGpuTypes("supported")`, which actually checks for drivers and CUDA runtime libraries before reporting a GPU type as available. On my system that correctly returns just `["vulkan"]`.

It may also be good to update the fallback logic. At the moment, it picks the first GPU type, and if it fails, jumps straight to CPU instead of trying the next one:

```js
const preferred = ["cuda", "metal", "vulkan"].find(g => gpuTypes.includes(g));
if (preferred) {
    try {
        llama = await getLlama({ gpu: preferred });
    } catch {
        llama = await getLlama({ gpu: false }); // skips vulkan
    }
}
```

Fixing the `getLlamaGpuTypes` call is probably enough, but looping through available types may be more resilient:

```js
const gpuTypes = await getLlamaGpuTypes("supported");
const gpuPriority = ["cuda", "metal", "vulkan"].filter(g => gpuTypes.includes(g));
let initialized = false;
for (const gpu of gpuPriority) {
    try {
        llama = await getLlama({ gpu, logLevel: LlamaLogLevel.error });
        initialized = true;
        break;
    } catch {
        process.stderr.write(`QMD Warning: ${gpu} failed to initialize.\n`);
    }
}
if (!initialized) {
    llama = await getLlama({ gpu: false, logLevel: LlamaLogLevel.error });
}
```

With either fix + `@node-llama-cpp/linux-x64-vulkan` as a dependency, embedding runs on GPU in about ~76s for 300 docs on my system, whereas on the CPU I'd have to wait ~15m. 

Happy to open a PR if it'd help.

**Environment:** Arch Linux, AMD RX 7900 XTX, AMD 5950x, ROCm + Vulkan, qmd 1.0.7, node-llama-cpp 3.15.1

### LINKED ISSUE #307: Add CPU-only mode to skip CUDA probe on non-GPU hosts
On a Linux CPU-only host (no NVIDIA GPU, no nvcc), qmd query attempts CUDA probe/build before falling back to CPU.

Observed stderr:

• CUDA Toolkit not found
• Failed to build llama.cpp with CUDA support
• then CPU fallback works.

Problem: noisy stderr can pollute structured automation (--json).

Requests:

1. Add first-class CPU-only option (e.g. --gpu cpu or QMD_GPU=cpu) that skips GPU probe entirely.
2. Keep --json mode stderr-clean for probe/build logs (or add --quiet).
3. Document recommended CPU-only config for Linux servers.

Environment:

• Linux x64
• Intel iGPU only (no NVIDIA)
• nvidia-smi absent
• nvcc absent

giladgd:
[`build: "autoAttempt"`](https://node-llama-cpp.withcat.ai/api/type-aliases/LlamaOptions#build) works similarly to the current behavior in `qmd` but is more efficient, faster, and allows the use of `NODE_LLAMA_CPP_GPU` env var to customize the default `gpu` used in `getLlama` when it's unset.

It should fix most of the issues users experience with CUDA not loading since `node-llama-cpp`'s internal backend resolution flow is more detailed and handles many edge cases.
It should also fix issues where CUDA is attempted to be used despite the system not supporting it (the current code doesn’t pass the required parameter to [`getLlamaGpuTypes`](https://node-llama-cpp.withcat.ai/api/functions/getLlamaGpuTypes) which is undefined behavior).

Fixes #307
Fixes #287 
Fixes #268 
Resolves #293
Resolves #266
Resolves #250
Resolves #246
Resolves #243
Resolves #240
Resolves #237
Partially remedies #222
Fixes #213
Fixes #194 
Fixes #185
Fixes #87 (the new `node-llama-cpp` version properly detects `glibc` on nixos now)
Fixes #59 (the new `node-llama-cpp` version contains a workaround for this bug in Bun)

### DIFF
--- a/src/collections.ts
+++ b/src/collections.ts
@@ -37,7 +37,6 @@ export interface Collection {
  */
 export interface CollectionConfig {
   global_context?: string;                    // Context applied to all collections
-  attempted_llama_build?: boolean;            // Whether llama build was already attempted
   collections: Record<string, Collection>;    // Collection name -> config
 }
 


tobi:
Thank you so much! really appreciate it @giladgd

PR #320, opened 2026-03-08
fix: add shell wrapper to resolve bun/node ABI mismatch

### LINKED ISSUE #319: bun global install fails when run with node: native module ABI mismatch (better-sqlite3)
## Problem

When installing via `bun install -g @tobilu/qmd`, the native addon `better-sqlite3` is compiled against bun's internal ABI (`NODE_MODULE_VERSION 141`). The CLI shebang is `#!/usr/bin/env node`, so on systems where `node` resolves to a standard Node.js binary (e.g. nvm-managed v22.x, `NODE_MODULE_VERSION 127`), every command fails with:

```
was compiled against a different Node.js version using NODE_MODULE_VERSION 141.
This version of Node.js requires NODE_MODULE_VERSION 127.
```

The README presents bun and npm as equivalent install options, so this is easy to hit.

**Workaround:** reinstall via npm instead of bun — `npm install -g @tobilu/qmd`. npm compiles native modules against the system node and the shebang works correctly.

## Reproduction

```sh
bun install -g @tobilu/qmd
qmd status   # fails with ERR_DLOPEN_FAILED if system node != bun's ABI version
```

## Proposed Fix

Replace the hardcoded `#!/usr/bin/env node` shebang with a shell wrapper that picks the right runtime based on how the package was installed. Since `dist/qmd.js` is the bin entry point, swap it for a small wrapper at `bin/qmd`:

```sh
#!/bin/sh
DIR="$(cd "$(dirname "$0")/.." && pwd)"
if command -v bun >/dev/null 2>&1; then
  exec bun "$DIR/dist/qmd.js" "$@"
else
  exec node "$DIR/dist/qmd.js" "$@"
fi
```

And update `package.json`:

```json
"bin": {
  "qmd": "bin/qmd"
}
```

This way bun-installed users run via bun (matching ABI), npm-installed users run via node. No breaking changes.

Happy to submit a PR with this if useful.

wolfkevin:
## Problem

When installed via \`bun install -g\`, \`better-sqlite3\` is compiled against bun's internal ABI (\`NODE_MODULE_VERSION 141\`). The hardcoded \`#!/usr/bin/env node\` shebang in \`dist/qmd.js\` then causes every command to fail on systems where \`node\` resolves to a standard Node.js binary (e.g. nvm-managed v22.x, module version 127):

\`\`\`
Error: The module '.../better_sqlite3.node'
was compiled against a different Node.js version using NODE_MODULE_VERSION 141.
This version of Node.js requires NODE_MODULE_VERSION 127.
\`\`\`

The README presents bun and npm as equivalent install options, so this is easy to hit. Fixes #319.

## Fix

Replaces the hardcoded shebang with a \`bin/qmd\` shell wrapper that detects the install method via path and picks the matching runtime:

\`\`\`sh
#!/bin/sh
DIR="$(cd "$(dirname "$0")/.." && pwd)"
case "$DIR" in
  */.bun/*) exec bun "$DIR/dist/qmd.js" "$@" ;;
  *) exec node "$DIR/dist/qmd.js" "$@" ;;
esac
\`\`\`

If the package is installed under \`.bun/\` (i.e. via \`bun install -g\`), it runs under bun — matching the ABI used to compile the native addon. Otherwise it falls back to node.

Also updates \`package.json\` to point \`bin.qmd\` at \`bin/qmd\` and adds \`bin/\` to the \`files\` array.

Note: this aligns with the direction already described in \`CLAUDE.md\` ("The \`qmd\` file is a shell script that runs compiled JS from \`dist/\`").

### DIFF
--- a/bin/qmd
+++ b/bin/qmd
@@ -1,7 +1,6 @@
 #!/bin/sh
 DIR="$(cd "$(dirname "$0")/.." && pwd)"
-if command -v bun >/dev/null 2>&1; then
-  exec bun "$DIR/dist/qmd.js" "$@"
-else
-  exec node "$DIR/dist/qmd.js" "$@"
-fi
+case "$DIR" in
+  */.bun/*) exec bun "$DIR/dist/qmd.js" "$@" ;;
+  *) exec node "$DIR/dist/qmd.js" "$@" ;;
+esac


wolfkevin:
Closing - superseded by #362 which merged a similar fix with an improvement (avoids the `$BUN_INSTALL` false-positive). The upstream `bin/qmd` now covers this.

PR #321, opened 2026-03-08
docs: add scheduled re-indexing guide (macOS launchd + Linux cron)

wolfkevin:
The README covers one-time setup (`qmd embed`) but doesn't explain how to keep the index fresh as notes grow and change over time.

Adds a "Keeping Your Index Fresh" section with:
- The `qmd update && qmd embed` pattern
- macOS launchd plist example (4-hour interval)
- Linux cron example

Placed before "Data Storage" so it sits naturally after the index maintenance commands.

### DIFF
--- a/docs/auto-update.md
+++ b/docs/auto-update.md
@@ -11,13 +11,6 @@ To automate this, schedule both commands to run periodically.
 
 ## macOS (launchd)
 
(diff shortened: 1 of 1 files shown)


wolfkevin:
Closing - upstream has moved significantly since this was opened and rebasing would be non-trivial. The scheduled re-indexing docs are still a gap but happy to re-open or re-submit if there's interest.

PR #325, opened 2026-03-08
feat: add remote LLM backend with averaged embedding optimization

olyashok:
## Summary

Adds an optional HTTP-based remote LLM backend (targeting llama.cpp servers) as an alternative to local node-llama-cpp, plus an embedding optimization that reduces query time significantly on large indexes.

**New features:**
- `src/llm-remote.ts`: `RemoteLLM` class — embed/rerank/generate via HTTP (llama.cpp compatible)
- `hybridQuery`, `vectorSearchQuery`, `structuredSearch` accept an optional `llm` override so callers can use a remote backend without changing global config
- Average all expanded query embeddings into a single vector → one sqlite-vec scan instead of N, cutting query time from ~47s to ~12s on a 25 GB index
- New CLI commands for remote mode; `--local` flag to force local node-llama-cpp
- `.qmd` directory resolution for DB and collection config paths
- `server/docker-compose.yml` for spinning up llama.cpp embed/rerank/generate services

**Why:** On large corpora (200k+ files) local embedding becomes a bottleneck. This lets QMD scale to server-side models while keeping local mode fully intact.

## Test plan

- [ ] Local mode (`--local`) still works unchanged
- [ ] Remote mode connects to llama.cpp server and returns correct results
- [ ] Averaged embeddings produce equivalent quality results to N-scan approach

---
🤖 AI-assisted (Claude) | Tested on local instance with ~25 GB index

### DIFF
--- a/server/docker-compose.yml
+++ b/server/docker-compose.yml
@@ -1,0 +1,68 @@
+name: qmd
+
+services:
+  qmd-embed:
+    image: ghcr.io/ggml-org/llama.cpp:server-cuda
+    container_name: qmd-embed
+    ports:
+      - "8081:8080"
+    volumes:
+      - qmd-models:/models
+    command: >
+      -m /models/embeddinggemma-300M-Q8_0.gguf
+      --embedding --pooling mean
+      --host 0.0.0.0 --port 8080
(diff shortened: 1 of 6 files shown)


tobi:

tobi:

PR #326, opened 2026-03-08
fix(search): match snake_case identifiers in BM25 queries

### LINKED ISSUE #305: BM25 search fails on snake_case identifiers (sanitizeFTS5Term strips underscores)
## Problem

Searching for Python/Ruby/Rust snake_case identifiers returns 0 results via BM25 search.

**Example**: documents contain `atomic_write_json` as a function name, but:

```
qmd search "atomic_write_json"    # → 0 results
qmd search "atomic write json"    # → 3 results, 88% score
```

## Root Cause

In `store.js`, `sanitizeFTS5Term` replaces non-alphanumeric characters with **empty string** instead of **space**:

```javascript
// Current (line ~1511):
function sanitizeFTS5Term(term) {
    return term.replace(/[^\p{L}\p{N}']/gu, '').toLowerCase();
}
```

This collapses `atomic_write_json` → `atomicwritejson` (single token).

However, at index time, FTS5's `unicode61` tokenizer correctly splits on `_`, storing `atomic`, `write`, `json` as separate tokens. The mismatch means the query can never match.

Additionally, `buildFTS5Query` splits only on `\s+` (whitespace), so even if `_` were preserved, it would still be treated as a single term.

## Fix (2-line change)

```diff
 function sanitizeFTS5Term(term) {
-    return term.replace(/[^\p{L}\p{N}']/gu, '').toLowerCase();
+    return term.replace(/[^\p{L}\p{N}']/gu, ' ').toLowerCase().trim();
 }
 function buildFTS5Query(query) {
-    const terms = query.split(/\s+/)
+    const terms = query.split(/[\s_]+/)
```

**Why this works**: replacing non-alnum with space (instead of empty) makes `atomic_write_json` → `atomic write json` → 3 AND terms, matching the FTS5 index. The `_` in the split pattern handles edge cases where sanitize alone doesn't split (e.g., already-sanitized input).

**No regressions**: normal multi-word queries (`"davinci assembler render"`) continue to work identically since spaces are already the split delimiter.

## Context

This is especially impactful for code indexing use cases where `project_code` collections contain Python/Ruby/Rust source files full of snake_case identifiers — the primary search vocabulary for developers.

## Environment

- qmd 1.0.7 (a0bd077aaf)
- macOS, Homebrew install

sonwr:
This fixes BM25 search for snake_case identifiers such as `atomic_write_json`.

What changed:
- replace non-alphanumeric characters with spaces instead of removing them during FTS5 term sanitization
- align the CLI-side sanitization logic with the store implementation
- add an end-to-end CLI test covering snake_case identifier search
- update structured search tests for the new tokenization behavior

Why:
SQLite FTS5 with the `unicode61` tokenizer splits `snake_case` into separate tokens at index time. The previous query sanitization removed underscores, turning `atomic_write_json` into `atomicwritejson`, which could not match the indexed tokens.

How I tested it:
- reproduced locally before the change:
  - `search "atomic_write_json"` returned `[]`
  - `search "atomic write json"` returned the expected document
- verified locally after the change:
  - `search "atomic_write_json"` returns the expected document
  - `search "parse_http_response"` returns the expected document
- ran targeted tests:
  - `test/structured-search.test.ts`
  - `test/cli.test.ts`

Fixes #305

### DIFF
--- a/src/qmd.ts
+++ b/src/qmd.ts
@@ -1720,8 +1720,9 @@ async function vectorIndex(model: string = DEFAULT_EMBED_MODEL, force: boolean =
 
 // Sanitize a term for FTS5: remove punctuation except apostrophes
 function sanitizeFTS5Term(term: string): string {
-  // Remove all non-alphanumeric except apostrophes (for contractions like "don't")
-  return term.replace(/[^\w']/g, '').trim();
+  // Replace punctuation with spaces so snake_case and similar identifiers
+  // are tokenized the same way as the FTS5 unicode61 index.
+  return term.replace(/[^\p{L}\p{N}']/gu, ' ').toLowerCase().trim();
 }
 
 // Build FTS5 query: phrase-aware with fallback to individual terms
--- a/src/store.ts
+++ b/src/store.ts
@@ -2032,7 +2032,7 @@ export function getTopLevelPathsWithoutContext(db: Database, collectionName: str
 // =============================================================================
 
 function sanitizeFTS5Term(term: string): string {
-  return term.replace(/[^\p{L}\p{N}']/gu, '').toLowerCase();
+  return term.replace(/[^\p{L}\p{N}']/gu, ' ').toLowerCase().trim();
 }
 
 /**
--- a/test/cli.test.ts
+++ b/test/cli.test.ts
@@ -591,6 +591,33 @@ describe("CLI Output Formats", () => {
       expect(stdout.toLowerCase()).toContain("api");
     }
   });
+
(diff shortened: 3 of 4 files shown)


sonwr:
Thanks. This change is small and still looks directionally fine, but the branch has gone stale and is now conflicting with the base branch. Rather than keep a conflicting PR open without active reviewer traction, I am closing this out here. If it still makes sense against current main later, I would prefer to reopen it as a fresh rebase or smaller follow-up.

PR #327, opened 2026-03-09
fix nix flake build and add helper script to maintain deps sha in flake

pkieltyka:
## Problem

The Nix package fails to build because bun install --frozen-lockfile runs during the build phase, which attempts to download dependencies from the internet. Nix builds run in a sandboxed environment with no network access, so every tarball download fails with ConnectionRefused / FailedToOpenSocket.

## Fix

The build is split into two derivations:
1. bunDeps — a fixed-output derivation (FOD) that runs bun install --frozen-lockfile --ignore-scripts. Fixed-output derivations are allowed network access by Nix because their output is verified against a content hash (outputHash). The --ignore-scripts flag is critical here — it skips native compilation steps (e.g. better-sqlite3's node-gyp build), which would produce outputs containing Nix store paths that break FOD content addressing.
2. qmd — the main derivation copies the pre-fetched node_modules from the FOD, makes them writable, patches shebangs for the Nix environment, and then runs bun install --frozen-lockfile. Since all packages are already present in node_modules, Bun skips downloading and only executes lifecycle scripts (native addon compilation etc.) — no network access needed.

## Maintaining the hash

Whenever bun.lock changes, the outputHash in flake.nix must be updated. A helper script is included for this:
bun run update-nix-hash

This temporarily sets the hash to a known-wrong value, runs nix build to get the correct hash from the mismatch error, and patches it back into flake.nix automatically.

## Changes
- flake.nix — split into two-phase build with FOD for dependency fetching
- scripts/update-nix-hash.sh — helper script to recompute the dependency hash
- package.json — added update-nix-hash script entry

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -6,8 +6,14 @@
     flake-utils.url = "github:numtide/flake-utils";
   };
 
-  outputs = { self, nixpkgs, flake-utils }:
-    flake-utils.lib.eachDefaultSystem (system:
+  outputs =
+    {
+      self,
+      nixpkgs,
+      flake-utils,
+    }:
+    flake-utils.lib.eachDefaultSystem (
+      system:
       let
         pkgs = nixpkgs.legacyPackages.${system};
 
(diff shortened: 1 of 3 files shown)


tobi:

tobi:

PR #333, opened 2026-03-09
fix(llm): add type cast for autoAttempt build option

### LINKED ISSUE #317: [Bug] v1.1.x / v2.0.x Build Error - TypeScript compilation failed
## Description

v1.1.5 fails to build with TypeScript compilation error:

```
src/llm.ts(548,27): error TS2769: No overload matches this call.
  Overload 1 of 2, lastBuild, gave the following error.
    Argument of type { build: string; logLevel: LlamaLogLevel; } is not assignable to parameter of type lastBuild.
  Overload 2 of 2, options?: LlamaOptions, gave the following error.
    Type autoAttempt is not assignable to type auto | never | forceRebuild | try | undefined.
```

## Environment

- OS: Linux
- Node.js version: v22.22.0
- npm version: 10.9.0

## Steps to Reproduce

```bash
cd qmd
git checkout v1.1.5
npm install
npm run build
```

## Expected Behavior

Build errors.

 should complete successfully without## Actual Behavior

Build fails with TypeScript error in src/llm.ts line 548.

## Workaround

Using v1.0.7 as a workaround.

JasonOA888:
## Problem

v1.1.5 fails to build with TypeScript compilation error:

```
src/llm.ts(548,27): error TS2769: No overload matches this call.
  Type autoAttempt is not assignable to type auto | never | forceRebuild | try | undefined
```

## Root Cause

The `"autoAttempt"` value is valid in node-llama-cpp runtime but not included in TypeScript type definitions.

## Solution

Cast to `any` to bypass type check while preserving runtime behavior.

## Alternative Considered

Could use `"try"` instead, but `"autoAttempt"` has better behavior (attempts build without forcing).

Fixes #317

### DIFF
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -546,8 +546,8 @@ export class LlamaCpp implements LLM {
   private async ensureLlama(): Promise<Llama> {
     if (!this.llama) {
       const llama = await getLlama({
-        // attempt to build
-        build: "autoAttempt",
+        // attempt to build (autoAttempt is valid but not in TS types yet)
+        build: "autoAttempt" as any,
         logLevel: LlamaLogLevel.error
       });
 
--- a/src/qmd.ts
+++ b/src/qmd.ts
@@ -64,7 +64,6 @@ import {
   type ExpandedQuery,
   type HybridQueryExplain,
   type StructuredSubSearch,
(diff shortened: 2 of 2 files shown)


tobi:

PR #334, opened 2026-03-09
fix(mcp): support Windows HOME and enable production mode

### LINKED ISSUE #277: MCP server returns 0 documents on Windows (HOME env var not supported)
## Bug Description

QMD MCP server returns 0 documents on Windows, even though the CLI works correctly and shows 417 indexed documents.

## Environment

- OS: Windows 11 Pro
- Node.js: v22.x
- QMD version: Latest from npm (`@tobilu/qmd`)
- Terminal: Git Bash / PowerShell

## Steps to Reproduce

1. Install QMD globally: `npm install -g @tobilu/qmd`
2. Index documents: `qmd index /path/to/docs`
3. Verify CLI works: `qmd search "test"` - Returns results correctly
4. Start MCP server: `qmd mcp` or configure in MCP client
5. Call search tool via MCP - Returns empty results

## Root Cause Analysis

Found two bugs in the code:

### Bug 1: `store.js` line 22 - Missing Windows HOME support

```javascript
const HOME = process.env.HOME || "/tmp";
```

On Windows, `process.env.HOME` is undefined (Windows uses `USERPROFILE`), causing `HOME` to default to `/tmp`. This makes the database path resolve to `/tmp/.cache/qmd/index.sqlite` instead of the correct Windows path `C:\Users\{username}\.cache\qmd\index.sqlite`.

**Fix:**
```javascript
const HOME = process.env.HOME || process.env.USERPROFILE || "/tmp";
```

### Bug 2: `mcp.js` - Missing `enableProductionMode()` call

The `startMcpServer()` function doesn't call `enableProductionMode()` before `createStore()`, which means the production database path isn't properly initialized.

**Fix:** Add `enableProductionMode()` call in `startMcpServer()` before creating the store.

## Current Workaround

Setting the `INDEX_PATH` environment variable forces the correct database path:

```json
{
  "command": "node",
  "args": ["path/to/wrapper.mjs"],
  "env": {
    "INDEX_PATH": "C:\\Users\\{username}\\.cache\\qmd\\index.sqlite"
  }
}
```

## Expected Behavior

MCP server should use the same database as the CLI and return the same 417 documents.

## Actual Behavior

MCP server returns 0 documents because it's looking for the database in `/tmp/.cache/qmd/index.sqlite` which doesn't exist.

JasonOA888:
## Problem

MCP server returns 0 documents on Windows even though CLI works correctly.

## Root Cause

Two bugs:

1. **store.ts:46** - `HOME` fallback only checks `process.env.HOME`, which is undefined on Windows (uses `USERPROFILE`). This causes database path to resolve to `/tmp/.cache/qmd/index.sqlite` instead of the correct Windows path.

2. **mcp.ts** - `startMcpServer()` and `startMcpHttpServer()` do not call `enableProductionMode()` before `createStore()`, causing the production database path to not be properly initialized.

## Solution

1. Change HOME fallback to: `process.env.HOME || process.env.USERPROFILE || "/tmp"`
2. Add `enableProductionMode()` call in both MCP server functions before `createStore()`

## Testing

- Verified the fix maintains backward compatibility on Unix systems
- Windows users should now see MCP server use the same database as CLI

Fixes #277

### DIFF
--- a/src/mcp.ts
+++ b/src/mcp.ts
@@ -22,6 +22,7 @@ import {
   addLineNumbers,
   structuredSearch,
   DEFAULT_MULTI_GET_MAX_BYTES,
+  enableProductionMode,
 } from "./store.js";
 import type { Store, StructuredSubSearch } from "./store.js";
 import { getCollection, getGlobalContext, getDefaultCollectionNames } from "./collections.js";
@@ -542,6 +543,7 @@ Intent-aware lex (C++ performance, not sports):
 // =============================================================================
 
 export async function startMcpServer(): Promise<void> {
+  enableProductionMode();
   const store = createStore();
   const server = createMcpServer(store);
   const transport = new StdioServerTransport();
@@ -563,6 +565,7 @@ export type HttpServerHandle = {
  * Binds to localhost only. Returns a handle for shutdown and port discovery.
  */
 export async function startMcpHttpServer(port: number, options?: { quiet?: boolean }): Promise<HttpServerHandle> {
+  enableProductionMode();
   const store = createStore();
 
   // Session map: each client gets its own McpServer + Transport pair (MCP spec requirement).
--- a/src/store.ts
+++ b/src/store.ts
@@ -43,7 +43,7 @@ import {
 // Configuration
 // =============================================================================
 
-const HOME = process.env.HOME || "/tmp";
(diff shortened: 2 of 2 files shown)


tobi:

PR #335, opened 2026-03-09
feat(update): add -c flag to update specific collections

### LINKED ISSUE #298: Update specific collection
Would it be possible to let the `-c` flag also apply to `qmd update`?

JasonOA888:
## Problem

The `-c/--collection` flag works with search commands but not with `qmd update`. Users with many collections have to update all of them even when they only want to re-index specific ones.

## Solution

Add collection filtering support to `updateCollections()` function:

```bash
qmd update -c notes              # Update only notes collection
qmd update -c notes -c docs      # Update multiple specific collections  
qmd update                       # Update all (existing behavior)
```

## Changes

- Add optional `collectionFilter` parameter to `updateCollections()`
- Filter collections list when `-c` flag is provided
- Show helpful message if no collections match the filter

Fixes #298

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -494,7 +494,7 @@ async function showStatus(): Promise<void> {
   closeDb();
 }
 
-async function updateCollections(): Promise<void> {
+async function updateCollections(collectionFilter?: string[]): Promise<void> {
   const db = getDb();
   const storeInstance = getStore();
   // Collections are defined in YAML; no duplicate cleanup needed.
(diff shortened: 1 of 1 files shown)


jtsternberg:
What's the status here? This would be really handy. My https://github.com/jtsternberg/snippets-cli has first-class for qmd, but when you add a snippet (e.g. `snip add --from-clipboard`), it (currently) has to update ALL collections which takes some time.

JasonOA888:
Thanks for the ping! I just rebased this PR onto the latest main. Should be ready for review now.

tobi:

PR #337, opened 2026-03-09
feat(llm): add QMD_FORCE_CPU env var for older GPUs

### LINKED ISSUE #299: Force cpu run on older nvidia architecture (e.g. Pascal)
When calling qmd query always run with CUDA, yet older architectures are not compatible:
```
[node-llama-cpp] CUDA error: the function requires an architectural feature absent from the device
```
To force it to run on cpu here is a workaround - run this (one time is enough):
```
CUDA_VISIBLE_DEVICES="" qmd query "test"
```
This is enough to build qmd with cpu only.

It would be good to have this small workaround in installation instructions.

JasonOA888:
## Problem

On older NVIDIA architectures (e.g., Pascal), `qmd query` fails with:

```
[node-llama-cpp] CUDA error: the function requires an architectural feature absent from the device
```

## Workaround (Before)

Users had to run:
```bash
CUDA_VISIBLE_DEVICES="" qmd query "test"
```

## Solution

Add `QMD_FORCE_CPU` environment variable:

```bash
export QMD_FORCE_CPU=1
qmd query "test"

# Or one-time:
QMD_FORCE_CPU=1 qmd query "test"
```

This disables GPU acceleration and forces CPU-only mode by passing `gpu: false` to `getLlama()`.

Fixes #299

### DIFF
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -545,10 +545,13 @@ export class LlamaCpp implements LLM {
    */
   private async ensureLlama(): Promise<Llama> {
     if (!this.llama) {
+      // QMD_FORCE_CPU=1 forces CPU-only mode (useful for older GPUs like Pascal)
+      const forceCpu = process.env.QMD_FORCE_CPU === "1" || process.env.QMD_FORCE_CPU === "true";
       const llama = await getLlama({
         // attempt to build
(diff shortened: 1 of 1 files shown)


tobi:

PR #332, opened 2026-03-09
fix(embed): display actual model name from QMD_EMBED_MODEL env var

JasonOA888:
## Problem

When using `QMD_EMBED_MODEL` to switch to a different embedding model, `qmd embed` still displays `Model: embeddinggemma` even though the correct model is downloaded and used.

```sh
export QMD_EMBED_MODEL="hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf"
qmd embed -f
# Output: Model: embeddinggemma  <-- incorrect
```

## Root Cause

Two different `DEFAULT_EMBED_MODEL` constants exist:
- `store.ts:47` - hardcoded display label `"embeddinggemma"`
- `llm.ts:196` - actual model URI (respects `QMD_EMBED_MODEL` env var)

`qmd.ts` was importing and using the wrong constant for display.

## Solution

1. Add `extractModelDisplayName()` helper to extract readable name from model URI
2. Use `DEFAULT_EMBED_MODEL_URI` (llm.ts) instead of `DEFAULT_EMBED_MODEL` (store.ts)
3. Apply `extractModelDisplayName()` when printing the model name

## Result

Now correctly shows:
```
Model: embeddinggemma-300M-Q8_0      # for default model
Model: Qwen3-Embedding-0.6B-Q8_0     # for custom models
```

Fixes #328

PR #348, opened 2026-03-10
fix(store): respect QMD_EMBED_MODEL env var in DEFAULT_EMBED_MODEL

### LINKED ISSUE #328: qmd embed displays hardcoded 'Model: embeddinggemma' when using QMD_EMBED_MODEL
## Bug

When using `QMD_EMBED_MODEL` to switch to a different embedding model (e.g. Qwen3-Embedding), `qmd embed` still displays `Model: embeddinggemma` even though the correct model is downloaded and used.

## Steps to reproduce

```sh
export QMD_EMBED_MODEL="hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf"
qmd embed -f
```

## Actual output

```
Force re-indexing: clearing all vectors...
Chunking 228 documents by token count...
Downloading to ~/.cache/qmd/models
✔ hf_Qwen_Qwen3-Embedding-0.6B-Q8_0.gguf downloaded 639.15MB in 11s
Downloaded to ~/.cache/qmd/models/hf_Qwen_Qwen3-Embedding-0.6B-Q8_0.gguf
Embedding 228 documents (3924 chunks, 6.5 MB)
196 documents split into multiple chunks
Model: embeddinggemma          <-- incorrect
```

The model is correctly downloaded and used (639MB Qwen3-Embedding), but the display name is wrong.

## Expected output

```
Model: Qwen3-Embedding-0.6B-Q8_0
```

## Cause

`store.ts:47` has a hardcoded `DEFAULT_EMBED_MODEL = "embeddinggemma"` which is passed to `vectorIndex()` at `qmd.ts:2847` as a display label. This value doesn't reflect the actual model loaded by `llm.ts` via `QMD_EMBED_MODEL`.

## Environment

- qmd v1.1.5
- macOS, M1 Pro 16GB

byheaven:
## Problem

When `QMD_EMBED_MODEL` is set to a custom embedding model (e.g. Qwen3-Embedding), **search queries use the wrong prompt template**, degrading search quality.

This is because `store.ts` hardcodes `DEFAULT_EMBED_MODEL = "embeddinggemma"`, and this value is passed through the call chain:

```
store.ts hybridQuery/vectorSearchQuery/structuredSearch
  → store.searchVec(query, DEFAULT_EMBED_MODEL, ...)
    → getEmbedding(query, "embeddinggemma", true, session)
      → formatQueryForEmbedding(query, "embeddinggemma")
```

In `formatQueryForEmbedding` (llm.ts:38), since `modelUri` is `"embeddinggemma"` (truthy), it does **not** fall through to `process.env.QMD_EMBED_MODEL`. Then `isQwen3EmbeddingModel("embeddinggemma")` returns `false`, so the query is formatted with the embeddinggemma template (`task: search result | query: ...`) instead of the Qwen3 instruct template.

This means documents are embedded with the correct Qwen3 format (because `llm.ts` reads the env var), but queries use the wrong format — causing a mismatch that hurts search quality.

## Solution

Make `store.ts` read `QMD_EMBED_MODEL` with the same fallback pattern as `llm.ts`:

```ts
// Before
export const DEFAULT_EMBED_MODEL = "embeddinggemma";

// After
export const DEFAULT_EMBED_MODEL = process.env.QMD_EMBED_MODEL ?? "embeddinggemma";
```

## Relation to #332

PR #332 fixes the **display** issue in `qmd.ts` (showing the wrong model name during `qmd embed`). This PR fixes the **functional** bug in `store.ts` where search queries use the wrong prompt template. The two fixes are complementary.

Fixes #328

Co-Authored-By: Oz <oz-agent@warp.dev>

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -44,7 +44,7 @@ import {
 // =============================================================================
 
 const HOME = process.env.HOME || "/tmp";
-export const DEFAULT_EMBED_MODEL = "embeddinggemma";
+export const DEFAULT_EMBED_MODEL = process.env.QMD_EMBED_MODEL ?? "embeddinggemma";
 export const DEFAULT_RERANK_MODEL = "ExpedientFalcon/qwen3-reranker:0.6b-q8_0";
 export const DEFAULT_QUERY_MODEL = "Qwen/Qwen3-1.7B";
 export const DEFAULT_GLOB = "**/*.md";


tobi:

PR #373, opened 2026-03-11
rename `file` param to `path` in get tool for LLM discoverability

### LINKED ISSUE #372: Rename `file` parameter to `path` in `get` tool for better LLM discoverability
## Problem

The `get` tool's primary parameter is named `file`, but it accepts a file path string (or docid). This causes LLMs to guess the wrong parameter name — `path` or `file_path` are the universal conventions for parameters that take file path strings (e.g., Claude's built-in `Read` tool uses `file_path`, most MCP tools use `path`).

In practice, when an LLM calls `get` without having fetched the full schema first, it will try `path` and fail — requiring an extra round-trip to discover the correct name is `file`.

## Proposed fix

Rename `file` → `path` in the `get` tool's `inputSchema`:

```ts
// src/mcp/server.ts

// Before
file: z.string().describe("File path or docid from search results (e.g., 'pages/meeting.md', '#abc123', or 'pages/meeting.md:100' to start at line 100)"),

// After
path: z.string().describe("File path or docid from search results (e.g., 'pages/meeting.md', '#abc123', or 'pages/meeting.md:100' to start at line 100)"),
```

Also update:
- Handler signature: `async ({ file, ... })` → `async ({ path, ... })`
- Error message: `` `Document not found: ${file}` `` → `` `Document not found: ${path}` ``
- `multi_get` error message: `Use 'qmd_get' with file="..."` → `Use 'qmd_get' with path="..."`

This is a minor breaking change for any client explicitly passing `{ file: "..." }` — but given the parameter name is actively misleading LLMs into errors, the fix is worth it. A CHANGELOG entry would cover it.

I'll open a PR with the change.

wolfkevin:
Fixes #372

## What

Renames the `file` parameter to `path` in the `get` tool's `inputSchema`.

## Why

`path` is the universal convention for parameters that accept file path strings - used by Claude's built-in `Read` tool (`file_path`), most MCP tools, and essentially every file-reading API. The name `file` reads as a file object, not a string path.

In practice, LLMs call `get` with `{ path: "..." }` on the first try and get a validation error, requiring an extra round-trip to discover the correct name is `file`. This is exactly what happened during a real session using this MCP - the error surfaced the issue and led to this fix.

## Changes

- `inputSchema`: `file` → `path`
- Handler destructuring: `{ file, ... }` → `{ path, ... }`  
- Error message: `` `Document not found: ${file}` `` → `` `Document not found: ${path}` ``
- `multi_get` skip message: `with file="..."` → `with path="..."`

Minor breaking change for clients explicitly passing `{ file: "..." }` - worth calling out in CHANGELOG.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -1,0 +1,1 @@
+placeholder
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -1,1 +1,807 @@
-placeholder
+/**
+ * QMD MCP Server - Model Context Protocol server for QMD
+ *
+ * Exposes QMD search and document retrieval as MCP tools and resources.
+ * Documents are accessible via qmd:// URIs.
+ *
+ * Follows MCP spec 2025-06-18 for proper response types.
+ */
+
+import { createServer, type IncomingMessage, type ServerResponse } from "node:http";
+import { randomUUID } from "node:crypto";
+import { fileURLToPath } from "url";
+import { McpServer, ResourceTemplate } from "@modelcontextprotocol/sdk/server/mcp.js";
+import { StdioServerTransport } from "@modelcontextprotocol/sdk/server/stdio.js";
+import { WebStandardStreamableHTTPServerTransport }
+  from "@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js";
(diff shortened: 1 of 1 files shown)


tobi:

Issue #363, opened 2026-03-11
Launcher script uses Bun instead of Node.js when $BUN_INSTALL is set, causing sqlite-vec failures


smc0210:
### Environment

- **QMD**: 2.0.1 (a0bd077aaf)
- **OS**: macOS 15 (Darwin 25.3.0), Apple M3 Max
- **Node.js**: via nvm
- **Bun**: 1.2.8 / 1.3.10 (both tested)
- **Installed via**: `npm install -g @tobilu/qmd`

### Problem

There are two related issues:

#### 1. Launcher script incorrectly selects Bun runtime

The launcher script at `/opt/homebrew/bin/qmd` checks for `$BUN_INSTALL` to decide the runtime:

```sh
if [ -f "$DIR/bun.lock" ] || [ -f "$DIR/bun.lockb" ] || [ -n "$BUN_INSTALL" ]; then
  exec bun "$DIR/dist/cli/qmd.js" "$@"
else
  exec node "$DIR/dist/cli/qmd.js" "$@"
fi
```

When QMD is installed via `npm install -g`, it should use Node.js. However, if `$BUN_INSTALL` is set (which is standard for any Bun user via `.zshrc`), the launcher forces Bun as the runtime regardless of how QMD was installed.

#### 2. Bun's built-in SQLite cannot load sqlite-vec extension

When running under Bun, any command that uses vector search (`query`, `vsearch`) crashes with:

```
SQLiteError: no such module: vec0
      at prepare (bun:sqlite:345:37)
```

Followed by a segfault:

```
panic(main thread): Segmentation fault at address 0xDFAE0
```

This occurs because `bun:sqlite` does not support loading native extensions like `sqlite-vec`, even when Homebrew SQLite (with extension support) is installed.

Tested on both Bun 1.2.8 and 1.3.10 — same result.

### Commands that work vs fail (under Bun)

| Command | Result |
|---|---|
| `qmd status` | ✅ Works |
| `qmd update` | ✅ Works |
| `qmd search "query"` | ✅ Works (BM25 only, no vec0) |
| `qmd embed` | ❌ sqlite-vec error |
| `qmd query "query"` | ❌ sqlite-vec error + segfault |
| `qmd vsearch "query"` | ❌ sqlite-vec error + segfault |

### Workaround

Replace the launcher script to force Node.js:

```sh
#!/bin/sh
exec node /opt/homebrew/lib/node_modules/@tobilu/qmd/dist/cli/qmd.js "$@"
```

[...]

Issue #238, opened 2026-02-21
Bun on macOS: sqlite-vec fails because Apple SQLite omits extension loading


mogsl:
## Problem

When running qmd under Bun on macOS, vector embedding fails at startup with:

```
sqlite-vec is not available. Vector operations require a SQLite build with extension loading support.
    at ensureVecTableInternal (store.js:608)
```

**Root cause:** Bun on macOS uses Apple's system SQLite for performance, but Apple's build sets `SQLITE_OMIT_LOAD_EXTENSION`, which disables `sqlite3_load_extension()`. The `sqlite-vec` npm packages are installed and the native `vec0.dylib` is present, but `db.loadExtension()` is rejected by the runtime.

## Environment

- qmd 1.0.7
- Bun 1.3.9
- macOS arm64 (Darwin 23.1.0)
- sqlite-vec 0.1.7 + sqlite-vec-darwin-arm64 (vec0.dylib present)
- Homebrew SQLite installed at `/opt/homebrew/opt/sqlite/lib/libsqlite3.dylib`

## Fix

Bun exposes `Database.setCustomSQLite(path)` to swap out Apple's restricted SQLite for a vanilla build that supports extension loading. The Bun branch in `db.ts` needs to call this before creating any Database instances.

Workaround patch applied to `dist/db.js`:

```js
if (isBun) {
    const bunSqlite = "bun:" + "sqlite";
    _Database = (await import(bunSqlite)).Database;
    // macOS: Apple's SQLite has SQLITE_OMIT_LOAD_EXTENSION; swap in Homebrew's vanilla build
    if (process.platform === "darwin") {
        const brewPrefix = process.env.BREW_PREFIX || (process.arch === "arm64" ? "/opt/homebrew" : "/usr/local");
        const customPath = `${brewPrefix}/opt/sqlite/lib/libsqlite3.dylib`;
        try {
            _Database.setCustomSQLite(customPath);
        } catch (e) {
            console.warn(`[qmd] failed to load Homebrew SQLite from ${customPath}: ${e.message}`);
        }
    }
    const { getLoadablePath } = await import("sqlite-vec");
    _sqliteVecLoad = (db) => db.loadExtension(getLoadablePath());
}
```

After applying this patch, `qmd embed` completes successfully and vector search works.

## References

- [Bun SQLite docs: `Database.setCustomSQLite()`](https://bun.sh/docs/api/sqlite#custom-sqlite-builds)
[...]

Issue #184, opened 2026-02-16
macOS: sqlite-vec fails to load — Apple SQLite disables extension loading


tongsh6:
## Environment

- **OS**: macOS (Intel x86_64)
- **Bun**: v1.3.9 (latest stable)
- **qmd**: @tobilu/qmd@1.0.0
- **sqlite-vec**: 0.1.7-alpha.2 (darwin-x64 dylib present and valid)

## Problem

`qmd query` fails with:

```
SQLiteError: no such module: vec0
    at prepare (bun:sqlite:331:37)
    at searchVec (src/store.ts:2069:25)
```

The root cause is that Bun on macOS uses **Apple's system SQLite**, which is compiled with extension loading disabled. When `db.loadExtension()` is called in `db.ts`, it throws:

```
This build of sqlite3 does not support dynamic extension loading
```

This error is caught silently in `initializeDatabase()` (setting `_sqliteVecAvailable = false`), but `searchVec()` in `hybridQuery()` is still called, causing the crash.

## Root Cause

`db.ts` line 17:
```ts
_sqliteVecLoad = (db: any) => db.loadExtension(getLoadablePath());
```

This works on Linux but fails on macOS because Apple's SQLite has `SQLITE_OMIT_LOAD_EXTENSION` set. Bun documents this and provides a workaround: `Database.setCustomSQLite()`.

See: https://bun.com/reference/bun/sqlite/Database/loadExtension

## Suggested Fix

In `db.ts`, before creating any Database instances, detect macOS and use Homebrew SQLite:

```ts
if (isBun) {
  const bunSqlite = await import("bun:sqlite");
  _Database = bunSqlite.Database;

  // macOS ships a SQLite that disables extension loading.
  // Use Homebrew SQLite if available so sqlite-vec can be loaded.
  if (process.platform === "darwin") {
    const brewPrefix = process.env.BREW_PREFIX ?? (
      process.arch === "arm64" ? "/opt/homebrew" : "/usr/local"
    );
    const brewSqlite = `${brewPrefix}/opt/sqlite/lib/libsqlite3.dylib`;
    try {
      const { statSync } = await import("node:fs");
      statSync(brewSqlite);
      _Database.setCustomSQLite(brewSqlite);
    } catch {
      // Homebrew SQLite not found; extension loading may fail
    }
  }

  const { getLoadablePath } = await import("sqlite-vec");
  _sqliteVecLoad = (db: any) => db.loadExtension(getLoadablePath());
}
```

[...]

Issue #161, opened 2026-02-12
Vector search fails silently when sqlite-vec extension loading fails in Bun environment


Guoaking:
## Environment
- **QMD version**: latest
- **Bun version**: 1.3.8
- **SQLite**: System 3.40.1 (supports extensions)
- **Platform**: Linux

## Description
When running QMD under Bun, `sqliteVec.load(db)` fails silently (no error thrown) because the underlying SQLite build doesn't support dynamic extension loading. This causes:

1. `vectors_vec` virtual table is not properly created (CREATE VIRTUAL TABLE succeeds but table is unusable)
2. Vector search (`vsearch`, `query`) times out or hangs
3. Only 1 corrupted row exists in `vectors_vec_chunks` instead of 346

## Expected Behavior
- `sqliteVec.load(db)` should throw a descriptive error if extension loading fails
- OR: code should verify the extension is actually loaded before using it

## Actual Behavior
- Extension loading fails silently
- `content_vectors` table works correctly (346 rows)
- `vectors_vec` table has only 1 corrupted row

## Reproduction Steps
1. Install QMD: `bun install -g https://github.com/tobi/qmd`
2. Run with Bun: `qmd embed --force`
3. Check vectors: `sqlite3 ~/.cache/qmd/index.sqlite "SELECT COUNT(*) FROM vectors_vec;"`
4. Run vector search: `qmd vsearch "test" -c memory-dir` ← Times out

## Workaround
Using `LD_PRELOAD=/usr/lib/x86_64-linux-gnu/libsqlite3.so.0 qmd ...` works because it forces Bun to use system SQLite.

## Suggested Fix
Add verification after `sqliteVec.load(db)`:

\`\`typescript
try {
  sqliteVec.load(db);
  // Verify extension actually loaded
  const testResult = db.prepare("SELECT vec_version()").get();
  if (!testResult?.version?.startsWith('v0')) {
    throw new Error("sqlite-vec extension not available, vector search disabled");
  }
} catch (err) {
  // Only handle specific "does not support" error
  if (!err.message.includes("does not support dynamic extension loading")) {
    throw err;
  }
}
\`\`

tobi:
Closing as part of the post-v2.5.1 backlog cleanup: this item is over two months old and has either been superseded by newer QMD releases or is too stale to keep actionable. If it still reproduces on v2.5.1 or newer, please open a fresh focused issue with current repro steps.

PR #377, opened 2026-03-11
fix(db): add macOS Homebrew SQLite support for Bun and restore actionable errors

serhii12:
## Problem

On macOS, `qmd embed` and `qmd query` fail under Bun with a cryptic error in `store.ts:770`:

```
sqlite-vec is not available. Vector operations require a SQLite build with extension loading support.
```

Two issues cause this:

1. **`bun:sqlite` uses Apple's system SQLite**, which is compiled with `SQLITE_OMIT_LOAD_EXTENSION` — native extensions like sqlite-vec cannot be loaded.
2. **The v2.0 refactor silently swallows the failure** in `initializeDatabase()`, setting `_sqliteVecAvailable = false` without any warning. The pre-2.0 code threw with actionable instructions ("Install Homebrew SQLite...") but this was replaced with a bare `catch {}`.

Reported in #363. Related: #238, #184, #161.

## Solution

**`src/db.ts`**:
- Call `Database.setCustomSQLite()` on macOS to swap Apple's SQLite with Homebrew's full-featured build. This is the pattern used by [sqlite-vec's own Bun example](https://github.com/asg017/sqlite-vec/blob/main/examples/simple-bun/demo.ts) and [Bun's docs](https://bun.com/docs/runtime/sqlite).
- Eagerly validate that extension loading works at init time (not at first query), so problems surface immediately.
- Throw with platform-specific fix instructions in `loadSqliteVec()` when the extension is unavailable.

**`src/store.ts`**:
- Log the error via `console.warn()` in `initializeDatabase()` instead of silently catching, so users see why vector search is disabled.

## Testing

1. **Node (unchanged path)**: `node dist/cli/qmd.js embed` works as before via better-sqlite3.
2. **Bun + macOS + Homebrew SQLite**: `bun dist/cli/qmd.js embed` works via `setCustomSQLite()`.
3. **Bun + macOS + no Homebrew**: User sees a clear warning with fix instructions. BM25 search still works.
4. **`qmd status` / `qmd search`**: Always work regardless of sqlite-vec availability.

## Configuration

Requires `brew install sqlite` on macOS when running under Bun. No change needed for Node users.

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -4,19 +4,51 @@
  * Provides a unified Database export that works under both Bun (bun:sqlite)
  * and Node.js (better-sqlite3). The APIs are nearly identical — the main
  * difference is the import path.
+ *
+ * On macOS, Apple's system SQLite is compiled with SQLITE_OMIT_LOAD_EXTENSION,
+ * which prevents loading native extensions like sqlite-vec. When running under
+ * Bun we call Database.setCustomSQLite() to swap in Homebrew's full-featured
+ * SQLite build before creating any database instances.
  */
 
 export const isBun = typeof globalThis.Bun !== "undefined";
 
 let _Database: any;
-let _sqliteVecLoad: (db: any) => void;
+let _sqliteVecLoad: ((db: any) => void) | null;
 
 if (isBun) {
   // Dynamic string prevents tsc from resolving bun:sqlite on Node.js builds
   const bunSqlite = "bun:" + "sqlite";
-  _Database = (await import(/* @vite-ignore */ bunSqlite)).Database;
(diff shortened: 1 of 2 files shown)


boozedog:
yes please!

eho:
FWIW, the fix works (MacOS 26.3.1) using the QMD SDK/library integrated directly within a Bun runtime. Vector searches and embeddings are executing without any sqlite-vec issues now. Thanks for fix!

tobi:
thank you

serhii12:
> thank you

Happy to help! Great project. Let me know if you ever want more hands on it.

PR #384, opened 2026-03-12
fix: allow hyphenated words in vec/hyde queries (#383)

### LINKED ISSUE #383: vec/hyde queries reject hyphenated words (kebab-case, compound adjectives) as negation syntax
# Issue: `validateSemanticQuery` rejects hyphenated/kebab-case words as negation syntax

## Title

vec/hyde queries reject hyphenated words (kebab-case, compound adjectives) as negation syntax

## Labels

bug, search

### Description

The `validateSemanticQuery` function in `src/store.ts` rejects any vec or hyde query containing a hyphenated word — including kebab-case identifiers (`node-llama-cpp`, `better-sqlite3`, `sqlite-vec`), compound adjectives (`real-time`, `long-lived`, `multi-client`), and multi-hyphen phrases (`state-of-the-art`, `end-to-end`, `copy-on-write`).

The validator is intended to catch lex-only negation syntax (`-term`, `-"phrase"`) in semantic queries, but the regex `/-\w/` matches any hyphen followed by a word character **anywhere** in the string — not just at token boundaries where negation actually occurs.

### Steps to Reproduce

```bash
# All of these fail with: "Negation (-term) is not supported in vec/hyde queries"
qmd query 'vec: how does the rate-limiter handle burst traffic'
qmd query 'vec: multi-client session architecture'
qmd query 'hyde: HTTP transport runs a single long-lived daemon shared across all clients'
qmd query 'vec: better-sqlite3 native module loading'
```

Via MCP:
```json
{
  "searches": [
    { "type": "vec", "query": "how does the rate-limiter handle burst traffic" }
  ]
}
```
Returns: `Structured search (vec): Negation (-term) is not supported in vec/hyde queries. Use lex for exclusions.`

### Expected

Queries with hyphenated words should be accepted. Only queries with negation syntax — a `-` at the start of the query or after whitespace — should be rejected.

### Actual

Any query containing a hyphen followed by a letter is rejected, regardless of position in the string.

### Root Cause

**`src/store.ts:2601`:**
```typescript
```

This is a character-level check, but negation is a token-level concept:
- `-redis` after whitespace → negation (exclude "redis")
- `long-lived` mid-word → hyphenated compound (not negation)

The regex cannot distinguish between these because it doesn't check what precedes the hyphen.

### Scope of Impact

Hyphenated words are extremely common in technical queries — exactly the kind of content QMD indexes:

**Kebab-case identifiers** (package names, CLI flags, CSS properties):
`node-llama-cpp`, `better-sqlite3`, `sqlite-vec`, `font-size`, `--no-verify`

**Compound adjectives**:
`real-time`, `long-lived`, `self-hosted`, `multi-client`, `cross-platform`, `non-blocking`, `single-threaded`

**Prepositional compounds**:
`in-memory`, `on-device`, `pre-built`, `write-ahead`, `out-of-band`

**Multi-hyphen phrases**:
`state-of-the-art`, `end-to-end`, `man-in-the-middle`, `copy-on-write`

**Short hyphenated terms**:
`e-commerce`, `A-B`, `re-index`, `co-located`

Any vec or hyde query containing these patterns is currently unusable.

### Proposed Fix

Anchor the negation check to token boundaries — start of string or after whitespace:

```diff
- if (/-\w/.test(query) || /-"/.test(query)) {
+ if (/(^|\s)-[\w"]/.test(query)) {
```

This also consolidates the two separate regex checks (`-\w` and `-"`) into a single pattern.

**Still correctly rejected** (negation syntax):
- `performance -sports` — negation after space
- `-redis connection pooling` — negation at start of query
- `-"exact phrase"` — negated quoted phrase
- `error handling -java -python` — multiple negations
- `  -term` — negation after leading whitespace
- `foo\t-bar` — negation after tab

**Now correctly accepted** (hyphenated words):
- `long-lived server shared across clients`
- `real-time voice processing pipeline`
- `state-of-the-art embedding models`
- `better-sqlite3 native module loading`
- `built-in vs add-on features`

### Environment

- QMD v2.0.1
- Affects both CLI (`qmd query`) and MCP server (`query` tool)

rymalia:
## Summary
- Tighten `validateSemanticQuery` regex from `/-\w/` to `/(^|\s)-[\w"]/` so negation is only detected at token boundaries
- Hyphenated/kebab-case words (`real-time`, `multi-client`, `better-sqlite3`, `state-of-the-art`) now pass validation
- Actual negation syntax (`-term`, `-"phrase"`) still correctly rejected
- Added 14 new test cases covering both directions

## Test plan
- [x] All 709 existing tests pass
- [x] 14 new `validateSemanticQuery` tests covering hyphenated acceptance + negation rejection
- [x] End-to-end CLI test: `qmd query 'vec: long-lived daemon'` passes validation

Fixes #383

🤖 Generated with my pal CC, [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -2597,8 +2597,9 @@ function buildFTS5Query(query: string): string | null {
  * Returns error message if invalid, null if valid.
  */
 export function validateSemanticQuery(query: string): string | null {
-  // Check for negation syntax
-  if (/-\w/.test(query) || /-"/.test(query)) {
+  // Check for negation syntax — only at token boundaries (start of string or after whitespace).
+  // Hyphenated words like "real-time" or "write-ahead" must not trigger this.
+  if (/(^|\s)-[\w"]/.test(query)) {
     return 'Negation (-term) is not supported in vec/hyde queries. Use lex for exclusions.';
   }
   return null;
--- a/test/structured-search.test.ts
+++ b/test/structured-search.test.ts
@@ -361,15 +361,71 @@ describe("lex query syntax", () => {
       expect(validateSemanticQuery("what is the CAP theorem")).toBeNull();
     });
 
-    test("rejects negation syntax", () => {
+    test("rejects negation at start of query", () => {
+      expect(validateSemanticQuery("-redis connection pooling")).toContain("Negation");
+    });
+
+    test("rejects negation after space", () => {
       expect(validateSemanticQuery("performance -sports")).toContain("Negation");
+    });
+
+    test("rejects negated quoted phrase", () => {
       expect(validateSemanticQuery('-"exact phrase"')).toContain("Negation");
     });
 
+    test("rejects multiple negations", () => {
+      expect(validateSemanticQuery("error handling -java -python")).toContain("Negation");
+    });
+
+    test("rejects negation after leading whitespace", () => {
+      expect(validateSemanticQuery("  -term at start")).toContain("Negation");
+    });
+
+    test("rejects negation after tab", () => {
+      expect(validateSemanticQuery("foo\t-bar")).toContain("Negation");
+    });
+
+    test("accepts hyphenated compound words", () => {
+      expect(validateSemanticQuery("long-lived server shared across clients")).toBeNull();
+      expect(validateSemanticQuery("real-time voice processing pipeline")).toBeNull();
+      expect(validateSemanticQuery("how does the rate-limiter handle burst traffic")).toBeNull();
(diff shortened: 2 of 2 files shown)


ferentchak:
Landed here from #383 — hit the same `validateSemanticQuery` rejection on v2.1.0 with hyphenated domain terms. Was going to submit a docs-only PR to clarify the current behavior, but the regex change here is trivial and actually resolves the bug instead of documenting around it. Would love to see this merged.

PR #400, opened 2026-03-13
feat(collections): add per-collection slugify option for path indexing

klokie:
## Summary

- Adds a `slugify` option to the `Collection` interface (default: `true` for backward compatibility)
- When `slugify: false`, filenames are stored as-is during indexing, preserving spaces, uppercase, and special characters
- Fixes path resolution for files like `BASS PLAYER Lesson 02.md` which were previously slugified to `bass-player-lesson-02.md`

### Motivation

QMD normalizes filenames to kebab-case during indexing via `handelize()`. This breaks path resolution for downstream tools (e.g. qmd-search-obsidian) when files have spaces or uppercase characters in their names. Making it configurable preserves backward compatibility while letting users opt in to original-path preservation.

### Usage

```yaml
# ~/.config/qmd/index.yml
collections:
  notes:
    path: /Users/me/Documents/notes
    pattern: "**/*.md"
    slugify: false   # preserve original filenames
```

## Test plan

- [x] New test: default behavior (`slugify: true`) still handleizes paths to kebab-case
- [x] New test: `slugify: false` preserves original filenames with spaces and uppercase
- [x] New test: search results return original filenames when `slugify: false`
- [x] All existing tests pass (450 tests across 7 test files)
- [ ] Manual test: index a collection with `slugify: false` containing files with spaces, verify `qmd search --json` returns correct paths

Made with [Cursor](https://cursor.com)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Per-collection `slugify` option (default: `true`). Set `slugify: false` in collection config to preserve original filenames with spaces and uppercase characters instead of converting to kebab-case during indexing
+
 ### Fixes
 
 - Sync stale `bun.lock` (`better-sqlite3` 11.x → 12.x). CI and release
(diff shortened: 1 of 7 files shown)


tobi:

PR #404, opened 2026-03-14
fix(store): preserve underscores in BM25 search terms

### LINKED ISSUE #305: BM25 search fails on snake_case identifiers (sanitizeFTS5Term strips underscores)

mvanhorn:
Fixes #305

## Summary

`sanitizeFTS5Term` stripped underscores from search terms, causing BM25 searches for snake_case identifiers to silently fail. `my_variable` became `myvariable`, matching nothing.

## Changes

- Add `_` to the preserved character set in `sanitizeFTS5Term` regex (`store.ts`)
- Export the function for testability
- Add 6 unit tests covering snake_case, contractions, punctuation, unicode

## Before / After

```
sanitizeFTS5Term("my_variable")
  Before: "myvariable"  (no BM25 match)
  After:  "my_variable" (correct match)
```

The CLI's copy of `sanitizeFTS5Term` (`cli/qmd.ts:1695`) already uses `\w` which preserves underscores - this aligns the store's version.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -2650,8 +2650,8 @@ export function getTopLevelPathsWithoutContext(db: Database, collectionName: str
 // FTS Search
 // =============================================================================
 
-function sanitizeFTS5Term(term: string): string {
-  return term.replace(/[^\p{L}\p{N}']/gu, '').toLowerCase();
+export function sanitizeFTS5Term(term: string): string {
+  return term.replace(/[^\p{L}\p{N}'_]/gu, '').toLowerCase();
 }
 
 /**
--- a/test/store.helpers.unit.test.ts
+++ b/test/store.helpers.unit.test.ts
@@ -16,6 +16,7 @@ import {
   isDocid,
   handelize,
   cleanupOrphanedVectors,
+  sanitizeFTS5Term,
 } from "../src/store";
 
 // =============================================================================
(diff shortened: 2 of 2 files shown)


mvanhorn:
Thanks for the quick merge!

PR #419, opened 2026-03-16
feat: added optional parameter to specify which host interface to listen on when using http transport

thrasher:
Treatment of the hard-coded "localhost" interface can vary by machine environment. This causes a problem specifically when trying to connect from a Docker service (ex: Open-WebUI via MCP) to QMD using host.docker.internal to localhost - it fails to connect. Specifying the 0.0.0.0 interface allows the connection, but alternatives should also be possible.

The --host parameter is optional, and defaults to the original behavior. Documentation and tests were added to support the change, while preserving the style. All tests pass.


dgjlindsay:
Just hit this exact issue — running QMD on Windows with Claude Code connecting from WSL2. Would love to see this merged. Clean change, tests included, backwards compatible. Happy to help test if useful.

tobi:

PR #420, opened 2026-03-17
fix: use per-call sessions for embeddings to prevent timeout on large collections

### LINKED ISSUE #410: Embedding fails after ~39,000 chunks on large documents
Environment: macOS, Apple M2 Max 32GB, Node v22.22.1
QMD version: latest
Database size: 544MB
Documents: 7 files, largest ~114MB markdown (split into 3 volumes)

Symptom: qmd embed consistently succeeds on ~39,000 chunks 
then reports 100,000+ failures regardless of document split size.
NODE_OPTIONS="--max-old-space-size=16384" set.

Steps to reproduce:
1. Index large markdown files (50MB+)
2. Run qmd embed
3. Observe consistent ~39,000 success then mass failure

debugerman:
The old design wrapped the entire embedding loop in a single LLM session with a 30-minute maxDuration cap. For collections exceeding ~39k chunks, the session timer would fire mid-loop and all remaining embeddings would fail with SessionReleasedError.

Each embed/embedBatch call now gets its own short-lived session with the default 10-minute timeout, which is more than sufficient for a single batch of ≤32 texts.

Closes #410

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -1323,125 +1323,137 @@ export async function generateEmbeddings(
   const totalDocs = docsToEmbed.length;
   const startTime = Date.now();
 
-  // Use store's LlamaCpp or global singleton, wrapped in a session
+  // Use store's LlamaCpp or global singleton.
   const llm = getLlm(store);
+  // Each embed call gets its own short-lived session so no single timer
+  // accumulates across the entire multi-batch job (see #410).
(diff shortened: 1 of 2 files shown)


tobi:
Closing — conflicts with #494 (which also touched generateEmbeddings) and Bun CI was failing. Happy to revisit if you rebase against main. Thanks!

PR #432, opened 2026-03-18
feat(mcp): expose plain query parameter for auto-expand search

fxstein:
## Problem

The SDK `search()` method accepts either `query` (plain string → auto-expanded by the trained model) or `queries` (pre-expanded typed sub-queries). The CLI `qmd query "text"` uses the `query` path, triggering the full pipeline: expansion → parallel BM25 + vector → RRF → reranking.

However, the MCP `query` tool only exposes the `queries` path via the `searches` parameter. There is no way to send a plain text query through MCP and get the full auto-expand pipeline.

The tool description Strategy table even says: *"Unknown vocabulary | Use a standalone natural-language query (no typed lines) so the server can auto-expand it"* — but the tool schema doesn't accept a standalone query.

## Fix

Add an optional `query` string parameter to the MCP `query` tool. When provided, route to `store.search({ query })` which calls `hybridQuery()` with auto-expansion. `query` and `searches` are mutually exclusive — providing both returns an error.

The `searches` parameter is made optional (was required). Validation ensures at least one of `query` or `searches` is provided.

Tool description Strategy table updated to recommend `query` as the default approach.

## Testing

Built and tested locally against a 7,791 document index:

| Query type | Latency | Results | Notes |
|-----------|---------|---------|-------|
| `query: "how does authentication work"` (cold) | 961ms | 5 | Expansion model loading |
| `query: "how does authentication work"` (warm) | 259ms | 5 | Cached — 3.7× faster |
| `query: "DEC-0054 sub-agent spawn protocol"` | 273ms | 5 | 93% top hit ✅ |
| `searches: [{type: "lex", query: "\"DEC-0054\""}]` | 197ms | 5 | Existing path, unchanged |
| `searches: [{type: "vec", query: "..."}]` | 622ms | 5 | Existing path, unchanged |

Existing `searches` parameter works identically to before — no breaking changes.

## Environment

- QMD: v2.0.1
- Platform: macOS (Apple Silicon)
- Node: v24.2.0

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -270,11 +270,12 @@ Combine types for best results. First sub-query gets 2× weight — put your str
 
 | Goal | Approach |
 |------|----------|
+| General search (recommended) | Use \`query\` parameter — auto-expands via trained model |
 | Know exact term/name | \`lex\` only |
 | Concept search | \`vec\` only |
 | Best recall | \`lex\` + \`vec\` |
 | Complex/nuanced | \`lex\` + \`vec\` + \`hyde\` |
-| Unknown vocabulary | Use a standalone natural-language query (no typed lines) so the server can auto-expand it |
+| Unknown vocabulary | Use \`query\` parameter with natural language so the server auto-expands it |
 
 ## Examples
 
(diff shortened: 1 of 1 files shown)


fxstein:
Rebased onto current main — picks up the new rerank param cleanly, both rerank and query now plumb through both code paths. Mergeable and ready when you have a moment.

fxstein:
The red `Bun (ubuntu-latest)` check is a pre-existing upstream-main regression since #537 merged on 2026-04-09 — unrelated to this PR. Fix in #602 (already green). Flagging so the red mark here doesn't distract from review.

tobi:

PR #438, opened 2026-03-19
feat: support remote OpenAI-compatible embedding via QMD_EMBED_API_URL

junmo-kim:
Adds support for routing embedding calls to a remote OpenAI-compatible API (e.g. Ollama, llama-server) via `QMD_EMBED_API_URL`, for environments where loading a local GGUF model isn't practical.

Submitted too early — closing to revise. Will reopen when it's properly ready.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


junmo-kim:
Sorry for the noise — submitted this too early. I'll get it properly ready and reopen when it's in better shape.

Issue #417, opened 2026-03-16
Lex search fails on hyphenated identifiers (e.g. DEC-0054, RFC-0011)


fxstein:
## Bug

`buildFTS5Query()` in `store.ts` mishandles hyphenated identifiers in lex queries. Bare `DEC-0054` is parsed as negation (`DEC` minus `0054`). Quoted `"DEC-0054"` gets sanitized to `dec0054` which doesn't match the FTS5 `unicode61` index tokens (`dec`, `0054`).

Related: #414 (same hyphen issue, but for vec/hyde queries)

## Reproduction

```
lex "DEC-0054"   → 0 results   (quoted phrase — sanitizer strips hyphen, becomes "dec0054")
lex DEC-0054     → 0 results   (bare term — parsed as "DEC" minus "0054", negation)
lex "dec 0054"   → works ✅    (workaround — but nobody types IDs this way)
```

## Expected

`"DEC-0054"` and `DEC-0054` should both find documents containing `DEC-0054`.

## Root Cause

Two issues in `buildFTS5Query()`:

1. **Bare terms:** `DEC-0054` hits the negation branch — hyphen is interpreted as `-term` operator.

2. **Quoted phrases:** `"DEC-0054"` goes through `sanitizeFTS5Term()` which strips all non-letter/non-number characters (`/[^\p{L}\p{N}']/gu`), producing `dec0054`. The FTS5 `unicode61` tokenizer splits hyphens at index time into separate tokens (`dec`, `0054`), but the query sanitizer concatenates them.

## Proposed Fix

- **Quoted phrases:** split on `[\s-]+` instead of `\s+`, so `"DEC-0054"` becomes phrase `"dec 0054"`
- **Bare terms:** when a term contains internal hyphens (not leading), split into multiple positive terms: `DEC-0054` → `"dec"* "0054"*`

This aligns query tokenization with `unicode61` index tokenization.

## Impact

Hyphenated identifiers are standard in technical knowledge bases: `RFC-0011`, `DEC-0054`, `GDE-0003`, `PIO-225`, `CVE-2024-1234`. All are unsearchable via lex without manually replacing hyphens with spaces.

## Environment

- QMD: v2.0.1
- Platform: macOS (Apple Silicon)
- Node: v24.2.0
- Installed via: npm

tobi:
[...]

Issue #390, opened 2026-03-12
vec/hyde queries reject valid text containing hyphens (false positive negation detection)


lunas-ai-lab:
## Problem

`validateSemanticQuery()` in `src/store.ts` rejects vec/hyde queries containing hyphens before words, treating them as negation operators. This causes false positives when queries contain CLI flags or hyphenated terms as part of natural language text.

**Example query:**
```json
{"type": "vec", "query": "autopilot script failed because claude -p command did not work"}
```

**Error:**
```
Structured search (vec): Negation (-term) is not supported in vec/hyde queries. Use lex for exclusions.
```

The `-p` here is a CLI flag reference, not a negation operator. Semantic search (embeddings) doesn't use negation syntax, so there's no reason to reject these queries.

## Current behavior

```typescript
export function validateSemanticQuery(query: string): string | null {
  if (/-\w/.test(query) || /-"/.test(query)) {
    return 'Negation (-term) is not supported in vec/hyde queries. Use lex for exclusions.';
  }
  return null;
}
```

The regex `/-\w/` matches any hyphen followed by a word character anywhere in the query — including mid-sentence CLI flags like `-p`, `-v`, compound words, etc.

## Proposed fix

Replace validation with sanitization — strip negation-like prefixes instead of rejecting the query. Embeddings ignore punctuation anyway, so removing the hyphen has no effect on search quality:

```typescript
export function sanitizeSemanticQuery(query: string): string {
  return query.replace(/(?:^|\s)-(?=\w|")/g, (match) => match.startsWith(' ') ? ' ' : '');
}
```

And in the caller, replace the throw with:
```typescript
} else if (search.type === 'vec' || search.type === 'hyde') {
  search.query = sanitizeSemanticQuery(search.query);
}
```

This way `"claude -p command"` → `"claude p command"` — search works, no error.

## Impact

This affects any MCP client (e.g. Claude Code) that generates vec/hyde queries with hyphens. Since the queries are LLM-generated, instructing the model to avoid hyphens is unreliable — the fix needs to be in code.

tobi:
Closing as part of the post-v2.5.1 backlog cleanup: this item is over two months old and has either been superseded by newer QMD releases or is too stale to keep actionable. If it still reproduces on v2.5.1 or newer, please open a fresh focused issue with current repro steps.

PR #463, opened 2026-03-25
Fix hyphenated tokens in FTS5 lex queries

### LINKED ISSUE #414: vec/hyde queries reject hyphenated compound words as negation operators
## Bug

`validateSemanticQuery()` in `store.js` (line ~1612) uses the regex `/-\w/` to detect negation syntax in vec/hyde queries. This regex matches any hyphen followed by a word character, which incorrectly rejects compound words like "multi-agent", "role-based", "chain-of-thought", etc.

## Reproduction

```json
{"searches": [{"type": "vec", "query": "multi-agent orchestration patterns"}]}
```

Returns: `Negation (-term) is not supported in vec/hyde queries. Use lex for exclusions.`

## Expected

The query should succeed — "multi-agent" is a compound word, not a negation operator.

## Root Cause

```javascript
// Current (buggy) — matches ANY hyphen + word char

// Fix — only match hyphens preceded by whitespace or at string start (actual negation syntax)
if (/(?:^|\s)-[\w"]/.test(query)) {
```

## Impact

Particularly acute for AI/ML knowledge bases where hyphenated terms are ubiquitous: multi-agent, few-shot, chain-of-thought, role-based, task-based, state-of-the-art, etc. Vec/hyde queries are essentially unusable for these domains.

## Version

v1.1.6, installed via bun

goldsr09:
## Summary

Fixes #414, #383, #384, #390, #417

Hyphenated tokens like `multi-agent`, `DEC-0054`, and `gpt-4` were broken in lex search because:

1. `sanitizeFTS5Term()` strips all non-alphanumeric characters, turning `multi-agent` into `multiagent`
2. The parser treated `-agent` as a negation prefix, so `multi-agent` became `multi NOT agent`

This PR:
- Adds `isHyphenatedToken()` to detect compound words with internal hyphens
- Adds `sanitizeHyphenatedTerm()` to split them into FTS5 phrase queries
- `multi-agent` now becomes `"multi agent"` (phrase match), not `"multiagent"*`
- Explicit negation (`-sports`) still works as before
- Negated hyphenated terms (`-multi-agent`) also work correctly

## Examples

| Input | Before | After |
|---|---|---|
| `multi-agent` | `"multiagent"*` | `"multi agent"` |
| `DEC-0054` | `"dec0054"*` | `"dec 0054"` |
| `gpt-4` | `"gpt4"*` | `"gpt 4"` |
| `multi-agent memory` | `"multi"* NOT "agent"* AND "memory"*` | `"multi agent" AND "memory"*` |
| `-sports` (negation) | `NOT "sports"*` | `NOT "sports"*` (unchanged) |

## Test plan

- [x] 8 new test cases covering hyphenated terms, mixed queries, negation disambiguation
- [x] All 12 existing `buildFTS5Query` tests pass (no regressions)
- [x] Full structured-search test suite passes

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -2655,19 +2655,45 @@ function sanitizeFTS5Term(term: string): string {
 }
 
 /**
+ * Check if a token is a hyphenated compound word (e.g., multi-agent, DEC-0054, gpt-4).
+ * Returns true if the token contains internal hyphens between word/digit characters.
+ */
+function isHyphenatedToken(token: string): boolean {
+  return /^[\p{L}\p{N}][\p{L}\p{N}'-]*-[\p{L}\p{N}][\p{L}\p{N}'-]*$/u.test(token);
+}
+
+/**
+ * Sanitize a hyphenated term into an FTS5 phrase by splitting on hyphens
+ * and sanitizing each part. Returns the parts joined by spaces for use
+ * inside FTS5 quotes: "multi agent" matches "multi-agent" in porter tokenizer.
+ */
+function sanitizeHyphenatedTerm(term: string): string {
+  return term.split('-').map(t => sanitizeFTS5Term(t)).filter(t => t).join(' ');
+}
+
+/**
  * Parse lex query syntax into FTS5 query.
  *
  * Supports:
  * - Quoted phrases: "exact phrase" → "exact phrase" (exact match)
  * - Negation: -term or -"phrase" → uses FTS5 NOT operator
+ * - Hyphenated tokens: multi-agent, DEC-0054, gpt-4 → treated as phrases
  * - Plain terms: term → "term"* (prefix match)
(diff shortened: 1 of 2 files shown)


rymalia:
Thanks for the lex fix — hyphenated tokens in FTS5 queries work correctly now.

However, this PR's `Fixes` line claims to resolve #383, #384, #390, and #414, which are all about `validateSemanticQuery()` rejecting vec/hyde queries containing hyphenated words. That's a different function (`store.ts:2910`) in a different code path — it wasn't modified by this PR.

The `/-\w/` regex in `validateSemanticQuery` is still unchanged in v2.1.0, so queries like `vec: "multi-agent orchestration"` or `vec: "AST-aware chunking"` are still rejected. #384 has the fix for that.

PR #476, opened 2026-03-27
feat: add --keep-models flag and QMD_KEEP_MODELS env var for MCP server

bagg-anon:
## Problem

When running `qmd mcp` or `qmd mcp --http` as a long-lived daemon, the current behaviour disposes **model weights** after 5 minutes of inactivity (`disposeModelsOnInactivity: true` hard-coded in `createStore`). This means the first query after an idle period has to reload the GGUF file from disk — which can take several seconds and causes a brief VRAM spike.

That trade-off makes sense for one-off CLI usage (you don't want to keep VRAM pinned after a single query), but not for a persistent MCP server where the whole point is low-latency responses.

The `LlamaCppConfig` already has `disposeModelsOnInactivity: false` as its default for exactly this reason (see the comment in `llm.ts`: *"Keeping models loaded avoids repeated VRAM thrash"*). We just had no way to opt into it from the outside.

## Solution

Add a `--keep-models` flag (and `QMD_KEEP_MODELS=1` env var) that keeps model **weights** in memory between queries while still releasing **contexts** (the lighter per-session VRAM objects) on inactivity.

### Changes

| File | Change |
|------|--------|
| `src/index.ts` | `StoreOptions` gains `keepModels?: boolean`; `createStore` reads `QMD_KEEP_MODELS` env var and passes the flag through to `LlamaCpp` |
| `src/mcp/server.ts` | `startMcpServer` and `startMcpHttpServer` accept `{ keepModels }` and forward it to `createStore` |
| `src/cli/qmd.ts` | `qmd mcp` gains `--keep-models` flag; it is also forwarded through the daemon spawn args; help text updated |

### Usage

```bash
# stdio transport (e.g. Claude Desktop, Cursor)
qmd mcp --keep-models

# HTTP foreground
qmd mcp --http --keep-models

# HTTP daemon (flag is forwarded to the child process)
qmd mcp --http --daemon --keep-models

# Via environment variable (handy for systemd units or Docker)
QMD_KEEP_MODELS=1 qmd mcp
```

### Why not flip the default?

The 5-minute timeout + `disposeModelsOnInactivity: true` is the right default for CLI use — most people run `qmd query` once and move on, and holding VRAM indefinitely would surprise them. The flag is an explicit opt-in for operators who know they are running a server.

## Testing

Verified by reading the logic path end-to-end:
- `--keep-models` → `keepModels = true` → `createStore({ keepModels: true })` → `disposeModelsOnInactivity: false` → model weights stay loaded, contexts are still released on inactivity timer
- `QMD_KEEP_MODELS=1` works the same way when the flag is absent
- Daemon spawn correctly appends `--keep-models` to child args so the flag survives `qmd mcp --http --daemon --keep-models`

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -2376,6 +2376,7 @@ function parseCLI() {
       http: { type: "boolean" },
       daemon: { type: "boolean" },
       port: { type: "string" },
+      "keep-models": { type: "boolean" },
     },
     allowPositionals: true,
     strict: false, // Allow unknown options to pass through
@@ -2565,6 +2566,7 @@ function showHelp(): void {
   console.log("  qmd multi-get <pattern>       - Batch fetch via glob or comma-separated list");
   console.log("  qmd skill show/install        - Show or install the packaged QMD skill");
   console.log("  qmd mcp                       - Start the MCP server (stdio transport for AI agents)");
+  console.log("    --keep-models               - Keep model weights in RAM between queries (good for servers)");
   console.log("");
   console.log("Collections & context:");
   console.log("  qmd collection add/list/remove/rename/show   - Manage indexed folders");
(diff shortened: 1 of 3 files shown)


keramblock:
works for me, via: 
```
&& sed -i 's/disposeModelsOnInactivity: true/disposeModelsOnInactivity: false/' /opt/qmd/lib/node_modules/@tobilu/qmd/dist/index.js \
    && grep -q 'disposeModelsOnInactivity: false' /opt/qmd/lib/node_modules/@tobilu/qmd/dist/index.js \
```
after installation In Dockerfile

tobi:

PR #486, opened 2026-03-29
Update outdated bun.lock file

surma:
Nix flake is currently broken because of an outdate bun.lock file


surma:
As far as I can tell, CI is failing not because of me. It seems 72f2dd1 is potentially the culprit.

tobi:
I've pushed a fix to this branch that restores the `toLowerCase()` call in `handelize()` (which was missing on main) and updated the tests to match. This should make CI green now, alongside your bun.lock update.

surma:
Closing this one in favor of #487

PR #487, opened 2026-03-29
Test nix flake builds in CI

surma:
To avoid stuff like #486 or #479, add a CI test that ensure the nix flake still works. WDYT?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,9 +18,13 @@
 ### Fixes
 
 - Fix paths in nix flake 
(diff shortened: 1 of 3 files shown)


surma:
(This new CI test will likely fail until #486 has landed :3)

tobi:
Merged the bun.lock and handelize fixes from #486 into this branch so CI for the new nix flake tests can pass.

surma:
@tobi well this was a journey. There was another test failure, some derivation hash instability on MacOS specifically... sorry this PR is a bit of a mess now, but CI’s passing. So that’s something.

PR #513, opened 2026-04-06
feat: Store original file paths alongside handelize'd paths

danmackinlay:
## Summary

- Adds a `source_path` column to the `documents` table that preserves original relative filesystem paths (e.g. `particle_filters.qmd` instead of the handelize'd `particle-filters.qmd`)
- Display code (`qmd ls`, search results, MCP responses) uses `COALESCE(source_path, path)` so pre-migration indexes degrade gracefully
- Lookup code (`findDocument`, `findDocuments`) resolves both handelize'd and original paths, so displayed paths are usable with `qmd get`
- `qmd status` reports files needing re-indexing to populate `source_path`

Extends the path handling work from #508. The handelize'd `path` column is kept for FTS tokenization (dashes are word boundaries, underscores are not) and dedup, while `source_path` is used for user-facing display.

## Test plan

- [x] 11 unit tests for source_path (display, lookup, fallback, migration idempotency)
- [x] 1 CLI integration test (index files with underscores, verify `ls`/`get`)
- [x] All 786 tests pass (LLM timeout tests are flaky under parallel contention but pass individually and with `--no-file-parallelism`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+### Changes
+
+- Store original filesystem paths (`source_path`) alongside handelize'd
+  paths. Search results, `qmd ls`, and MCP responses now display the
+  real filename (preserving underscores, case, etc.) instead of the
+  normalized version. Lookups accept both original and handelize'd paths.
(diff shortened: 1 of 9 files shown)


tobi:

PR #514, opened 2026-04-06
feat(cli): add qmd watch for auto-reindexing on file changes

mvanhorn:
## Summary

Adds `qmd watch [--embed]` that monitors all configured collection directories for file changes and auto-reindexes affected collections with 500ms debouncing.

## Why this matters

QMD requires manual `qmd update` to reindex after file changes. When editing notes throughout the day, search returns stale results until you remember to reindex.

| Source | Evidence | Signal |
|--------|----------|--------|
| [CK](https://github.com/BeaconBay/ck) | Auto-updates index on file changes | Direct competitor, 1.5K stars |
| [Khoj](https://github.com/khoj-ai/khoj) | Auto-syncs data sources on schedule | 34K stars |
| Dogfooding | After `collection add` + editing files, search returned stale results | First-run friction |

## Changes

- New `watchCollections()` function in `src/cli/qmd.ts` using Node.js `fs.watch` with recursive mode
- `watch` case in the CLI switch statement
- `--embed` option in `parseCLI()` to auto-trigger embedding after updates
- Help text updated
- Changelog entry added

Implementation uses `fs.watch` (event-based, macOS FSEvents / Linux inotify) with per-collection debouncing. No new dependencies. Reuses existing `reindexCollection()` for the actual indexing work.

By default, `watch` only reindexes (fast). Pass `--embed` to also regenerate vector embeddings after each update (loads models into VRAM).

## Demo

qmd watch demo

## Testing

- All 91 existing CLI tests pass
- Manual testing: started `qmd watch`, created a new `.md` file in a collection directory, observed auto-reindex output within 1 second

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -5,7 +5,7 @@ import { execSync, spawn as nodeSpawn } from "child_process";
 import { fileURLToPath } from "url";
 import { basename, dirname, join as pathJoin, relative as relativePath, resolve as pathResolve } from "path";
 import { parseArgs } from "util";
-import { readFileSync, readdirSync, realpathSync, statSync, existsSync, unlinkSync, writeFileSync, openSync, closeSync, mkdirSync, lstatSync, rmSync, symlinkSync, readlinkSync, copyFileSync } from "fs";
+import { readFileSync, readdirSync, realpathSync, statSync, existsSync, unlinkSync, writeFileSync, openSync, closeSync, mkdirSync, lstatSync, rmSync, symlinkSync, readlinkSync, copyFileSync, watch as fsWatch } from "fs";
 import { createInterface } from "readline/promises";
 import {
   getPwd,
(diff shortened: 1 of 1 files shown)


mvanhorn:
Rebased onto main -- b292fb5. Still relevant: there's no `watch` command on `main`.

Two conflicts worth noting. The `fs` import list had diverged on both sides, so it's merged rather than replaced -- `readdirSync`/`copyFileSync` from `main` plus `watch as fsWatch`.

The other one mattered more: `main` replaced the old `disposeDefaultLlamaCpp()` + `process.exit(0)` teardown with `finishSuccessfulCliCommand()`. This branch had excluded `watch` from that path because the watcher is long-running and must not exit. I kept `main`'s `finishSuccessfulCliCommand` and carried the `watch` exclusion onto it, so the new teardown still runs for every other command.

`qmd --help` lists the command, and `test/cli-exit-lifecycle.test.ts` is 6 pass / 0 fail.

yev9en:
Strong +1 from a data point we hit in production-style use (2.8.3,
linux-arm64): without auto-reindexing, keyword search doesn't just return
stale results — it silently *lies* about misses, which is much worse:

- Append a line with a unique token to an indexed file → `qmd search
  <token>` → "No results found." **exit 0**, no stderr, no hint the index
  is behind. The user concludes the note doesn't exist.
- Delete a file → `qmd search` keeps returning it at full score, and
  `qmd get` serves its body — until a manual `qmd update` runs.
- `qmd status` shows "Updated: 3m ago" and `qmd ls` shows the *index*
  mtime/size, so nothing surfaces the discrepancy.

So this isn't only a convenience feature; paired with the silent-failure
behavior it makes "No results found" an unreliable answer. Even a much
cheaper stopgap than a watcher would help a lot:

- a dirty-file check at query time (stat mtimes vs. index rows), with a
  stderr warning like "index stale — run qmd update", or
- an opt-in `qmd update --if-dirty` that scripts could call before search.

Full repro steps available if useful.

PR #515, opened 2026-04-06
feat(cli): add qmd similar for finding related documents

mvanhorn:
## Summary

Adds `qmd similar <path|#docid>` that finds semantically similar documents using existing embedding cosine similarity in sqlite-vec. Also exposed as an MCP `similar` tool and `store.findSimilarByEmbedding()` SDK method.

## Why this matters

QMD generates embeddings for every document but only uses them reactively (when you search). There's no way to ask "what documents are similar to this one?" Reor and Obsidian Smart Connections both build their UX around document similarity - it's one of the most-valued features in the note-taking space.

| Source | Evidence | Signal |
|--------|----------|--------|
| [Reor](https://github.com/reorproject/reor) | Auto-links related notes via embeddings | 8.5K stars, core feature |
| [Smart Connections](https://github.com/brianpetro/obsidian-smart-connections) | "Find similar" is primary interaction | 4.8K stars, core feature |
| QMD architecture | sqlite-vec already has `vec_distance_cosine()`, embeddings stored per doc | Zero new infrastructure |

The embedding infrastructure already exists. This is one SQL query on top of existing data.

## Changes

- New `findSimilarByEmbedding()` function in `src/store.ts` using the same two-step vec0 query pattern as `searchVec` (avoids sqlite-vec JOIN hangs per PR #23)
- `similar` case in CLI switch with path/docid argument
- MCP `similar` tool with path, n, collection parameters
- SDK `findSimilarByEmbedding()` method on QMDStore interface
- Help text and changelog entry

## Simulated Demo

qmd similar demo

## Testing

- All 91 existing CLI tests pass
- Build succeeds with no type errors

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -22,6 +22,9 @@ import {
   renameCollection,
   findSimilarFiles,
   findDocument,
+  findSimilarByEmbedding,
+  findDocumentByDocid,
+  isDocid,
   matchFilesByGlob,
   getHashesNeedingEmbedding,
   clearAllEmbeddings,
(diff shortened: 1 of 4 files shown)


mvanhorn:
Rebased onto main -- 28ebe99. Still relevant: `main` has `findSimilarFiles`, but that's fuzzy filename matching for "did you mean" suggestions -- there's no semantic document similarity and no `similar` command.

Both conflicts were additive merges rather than either/or. The store import list needed `findDocument` (which `main` uses) alongside the `findSimilarByEmbedding` / `findDocumentByDocid` / `isDocid` this branch needs, and the help text keeps `main`'s newer `qmd get` line with the `similar` line added above it. `isDocid` and `findDocumentByDocid` both still exist on `main`, so nothing else moved.

`qmd --help` lists the command, and the store tests pass: `store.helpers.unit` 35/0, `store-paths` + `store-cjk-fts` 49/0.

PR #525, opened 2026-04-07
fix(store): align default db home resolution with config

kuishou68:
## Summary
- align the default database home resolution path with the configured home directory
- keep store path resolution consistent with user configuration
- avoid surprising mismatches between configured home and default store location


tobi:

PR #527, opened 2026-04-07
fix: honor active embed model in query vector lookups

### LINKED ISSUE #497: QMD query mode ignores custom QMD_EMBED_MODEL in store path and causes 768/1024 dimension mismatch
# QMD issue draft — `query` mode dimension mismatch with custom `QMD_EMBED_MODEL`

## Suggested title

QMD query mode ignores custom `QMD_EMBED_MODEL` in store path and causes 768/1024 dimension mismatch

## Repository

- Target: `tobi/qmd`

## Summary

When indexing documents with a custom multilingual embedding model via:

```sh
export QMD_EMBED_MODEL="hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf"
```

embedding completes successfully and `qmd search` works as expected.

However, `qmd query` fails with a vector dimension mismatch:

```text
SqliteError: Dimension mismatch for query vector for the "embedding" column.
Expected 768 dimensions but received 1024.
```

This appears to happen because the query-time vector lookup path in `dist/store.js` still hardcodes `DEFAULT_EMBED_MODEL = "embeddinggemma"` and passes that into `searchVec()`, even though `dist/llm.js` correctly respects `QMD_EMBED_MODEL`.

## Environment

- QMD version: `2.0.1`
- Install method: global install via Bun (`bun install -g @tobilu/qmd`)
- OS: macOS
- Node: `v24.14.1`
- Embedding model override:

```sh
QMD_EMBED_MODEL="hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf"
```

## Steps to reproduce

1. Install QMD 2.0.1
2. Set:

```sh
export QMD_EMBED_MODEL="hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf"
```

3. Create / index a collection of markdown files
4. Force re-embed:

```sh
qmd embed -f
```

5. Run a normal search:

```sh
qmd search "QMD 統一知識庫搜尋規劃" --json -n 5
```

This returns valid results.

6. Run query mode:

```sh
qmd query "QMD 統一知識庫搜尋規劃" --json -n 5
```

or other Chinese-language queries such as:

```sh
qmd query "lossless-claw QMD 安裝順序" --json -n 5
qmd query "中文語料 embedding 決策 Qwen3-Embedding" --json -n 5
```

## Expected behavior

`qmd query` should use the same effective embedding model assumptions as the index that was built with `QMD_EMBED_MODEL`, and should return results instead of failing.

## Actual behavior

`qmd query` fails after query expansion / embedding with:

```text
SqliteError: Dimension mismatch for query vector for the "embedding" column.
Expected 768 dimensions but received 1024.
```

This suggests the query pipeline is using an embedding/vector assumption that does not match the indexed vectors.

### `dist/llm.js`

This file appears to respect the env var:

```js
const DEFAULT_EMBED_MODEL = process.env.QMD_EMBED_MODEL ?? "hf:ggml-org/embeddinggemma-300M-GGUF/embeddinggemma-300M-Q8_0.gguf";
```

### `dist/store.js`

This file defines a separate hardcoded constant:

```js
export const DEFAULT_EMBED_MODEL = "embeddinggemma";
```

And query-related paths pass that constant into `searchVec()`:

```js
const vecResults = await store.searchVec(vecQueries[i].text, DEFAULT_EMBED_MODEL, 20, collection, undefined, embedding);
```

```js
const vecResults = await store.searchVec(q, DEFAULT_EMBED_MODEL, limit, collection);
```

```js
const vecResults = await store.searchVec(vecSearches[i].query, DEFAULT_EMBED_MODEL, 20, coll, undefined, embedding);
```

This makes it look like query-time vector lookup may still assume `embeddinggemma` even when the actual index was built using `Qwen3-Embedding`.

## Additional observations

- `qmd search` still works for title / keyword queries
- `qmd query` consistently fails on this setup
- There is also a related issue about `DEFAULT_EMBED_MODEL` hardcoding / metadata drift: `#447`
- This may be more than metadata drift; it appears to affect runtime query behavior

## Suggested fix direction

- Avoid hardcoding `embeddinggemma` inside `store.js`
- Ensure query / structured / hybrid vector lookup uses the same effective embedding model as the one used during indexing
- Prefer propagating the active embed model (or embed model URI) from `llm.js` / session state into `searchVec()`
- At minimum, ensure internal model labels and vector dimension expectations stay aligned with `QMD_EMBED_MODEL`

## Workaround

For now, avoid `qmd query` when indexing with `QMD_EMBED_MODEL=Qwen3-Embedding...` and use:

```sh
qmd search ...
```

instead.

kuishou68:
## Summary
- use the active llm embed model for query-time vector lookups instead of the hardcoded default alias
- keep hybrid and vector-only query paths aligned with the model used to build embeddings
- add regression coverage for vector and hybrid query routing

Closes #497

## Testing
- corepack pnpm exec vitest run test/store.test.ts -t "vectorSearchQuery uses the active llm embed model for vector lookups" --reporter=dot
- corepack pnpm exec vitest run test/store.test.ts -t "hybridQuery uses the active llm embed model for precomputed vector lookups" --reporter=dot
- corepack pnpm build

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -3991,6 +3991,7 @@ export async function hybridQuery(
 
     // Batch embed all vector queries in a single call
     const llm = getLlm(store);
+    const embedModel = llm.embedModelName;
     const textsToEmbed = vecQueries.map(q => formatQueryForEmbedding(q.text, llm.embedModelName));
     hooks?.onEmbedStart?.(textsToEmbed.length);
     const embedStart = Date.now();
@@ -4003,7 +4004,7 @@ export async function hybridQuery(
       if (!embedding) continue;
 
       const vecResults = await store.searchVec(
-        vecQueries[i]!.text, DEFAULT_EMBED_MODEL, 20, collection,
+        vecQueries[i]!.text, embedModel, 20, collection,
         undefined, embedding
       );
       if (vecResults.length > 0) {
@@ -4233,10 +4234,11 @@ export async function vectorSearchQuery(
   options?.hooks?.onExpand?.(query, vecExpanded, Date.now() - expandStart);
 
   // Run original + vec/hyde expanded through vector, sequentially — concurrent embed() hangs
+  const embedModel = getLlm(store).embedModelName;
   const queryTexts = [query, ...vecExpanded.map(q => q.query)];
   const allResults = new Map<string, VectorSearchResult>();
   for (const q of queryTexts) {
-    const vecResults = await store.searchVec(q, DEFAULT_EMBED_MODEL, limit, collection);
+    const vecResults = await store.searchVec(q, embedModel, limit, collection);
     for (const r of vecResults) {
       const existing = allResults.get(r.filepath);
       if (!existing || r.score > existing.score) {
@@ -4374,6 +4376,7 @@ export async function structuredSearch(
     );
     if (vecSearches.length > 0) {
       const llm = getLlm(store);
+      const embedModel = llm.embedModelName;
       const textsToEmbed = vecSearches.map(s => formatQueryForEmbedding(s.query, llm.embedModelName));
       hooks?.onEmbedStart?.(textsToEmbed.length);
       const embedStart = Date.now();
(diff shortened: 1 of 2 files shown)


tobi:

Issue #488, opened 2026-04-01
Bug: Incorrect path handling on Windows when QMD_CONFIG_DIR is set


axulonash:
Bug: Incorrect path handling on Windows when QMD_CONFIG_DIR is set

When using qmd as an MCP server on Windows with QMD_CONFIG_DIR environment variable set, the tool incorrectly concatenates paths, resulting in an invalid database path.

## Steps to Reproduce

1. Set QMD_CONFIG_DIR environment variable to a Windows path (e.g., C:\Users\axulo\.config\qmd)
2. Run qmd MCP server (e.g., via OpenCode or similar tool that uses qmd as MCP)
3. Attempt to index or search documents in a working directory (e.g., C:\Users\axulo\Documents\ppttest)

## Expected Behavior
qmd should correctly identify the working directory and .mnemonics folder without path concatenation issues.

## Actual Behavior
qmd attempts to create/access database at an invalid path:
- Expected: Uses working directory (e.g., C:\Users\axulo\Documents\ppttest\.mnemonics)
- Actual: Tries to create path like C:\Users\axulo\.config\qmd\c:\-users\axulo\-documents

## Error Message
Failed to create database: Error: ENOENT: no such file or directory, mkdir 'C:\Users\axulo\.config\qmd\c:\-users\axulo\-documents'

## Environment
- OS: Windows
- qmd version: Latest (via npm @tobilu/qmd)
- Working Directory: C:\Users\axulo\Documents\ppttest
- QMD_CONFIG_DIR: C:\Users\axulo\.config\qmd

## Additional Context
This appears to be a path parsing bug where qmd is incorrectly combining the QMD_CONFIG_DIR value with the Windows-style working directory path, causing the backslashes to be treated as escape characters or literal characters rather than path separators.

## Suggested Fix
Ensure proper path normalization when combining QMD_CONFIG_DIR with the working directory path on Windows. The path joining logic should correctly handle both Unix-style and Windows-style paths.

tobi:
Closing as stale — no updates in 3+ months. Please reopen if this is still relevant on current main.

PR #528, opened 2026-04-07
fix: normalize Windows path index names

kuishou68:
## Summary\n- normalize path-like index names in one helper shared by CLI and config selection\n- sanitize Windows absolute paths when deriving config filenames under QMD_CONFIG_DIR\n- add regression coverage for Windows path handling\n\nCloses #488.


tobi:

PR #531, opened 2026-04-07
fix: respect explicit QMD_LLAMA_GPU backends

kuishou68:
## Summary
- preserve explicit QMD_LLAMA_GPU backends like metal, cuda, and vulkan
- keep CPU-only override values mapped to false
- add unit coverage for the GPU mode parser

Closes #529.


tobi:

PR #533, opened 2026-04-07
fix: include line in CLI --json search output

### LINKED ISSUE #505: Include line in --json search output
# Include `line` in `--json` search output

## Problem

The CLI formatter includes line numbers in search results (e.g. `qmd://blog/notebook/database_vector.qmd:2`), but the `--json` formatter omits them. The line number is computed from `chunkPos` via `extractSnippet` in the CLI path (`cli/qmd.js:1562`) but isn't passed through to `searchResultsToJson` in `formatter.js`.

This makes it harder to build editor integrations. For example, VS Code's `code --goto file:line` works great with the CLI output, but parsing coloured terminal output is fragile. A `--json` workflow like:

```bash
qmd query "vector database" --json | jq -r '.[0] | "\(.file):\(.line)"' | xargs code --goto
```

would be much more robust, but currently `.line` is absent from the JSON.

## Suggested fix

In `cli/formatter.js`, `searchResultsToJson` (around line 53), the result object is built as:

```js
return {
    docid: `#${row.docid}`,
    score: Math.round(row.score * 100) / 100,
    file: row.displayPath,
    title: row.title,
    ...(row.context && { context: row.context }),
    ...(body && { body }),
    ...(snippet && { snippet }),
};
```

Adding the line number would be something like:

```js
const { line, snippet: snip } = extractSnippet(bodyStr, query, 300, row.chunkPos, undefined, opts.intent);
return {
    docid: `#${row.docid}`,
    score: Math.round(row.score * 100) / 100,
    file: row.displayPath,
    line,
    title: row.title,
    ...(row.context && { context: row.context }),
    ...(body && { body }),
    ...(snip && !opts.full && { snippet: snip }),
};
```

The `extractSnippet` function is already imported in `qmd.js`; it may need to be made available to `formatter.js` as well, or the line could be pre-computed and passed through on the result object.

## Use case

Building shell/editor integrations that open search results at the matching line — VS Code (`code --goto`), Vim (`:e +line file`), Emacs, etc.

rymalia:
## Summary

- `--json` search output now includes a `line` field (1-indexed) pointing to the best-matching line in the document, enabling editor integrations like `code --goto file:line`

## Problem

#505 requested a `line` field in `--json` output for editor integrations. #506 was merged to fix it, but the fix landed in `searchResultsToJson()` in `formatter.ts` — a function with **no callers**. The CLI's actual JSON output is generated by `outputResults()` in `qmd.ts`, which has its own inline formatting. That inline path destructured only `.snippet` from `extractSnippet()`, discarding the `.line` property.

## Fix

Extract the full `SnippetResult` from `extractSnippet()` and spread `line` into the JSON output object — matching what the CLI display path already does for terminal output.

**Before:**
```json
{"file": "qmd://blog/post.md", "snippet": "@@ -42,4 @@ ..."}
```

**After:**
```json
{"file": "qmd://blog/post.md", "line": 42, "snippet": "@@ -42,4 @@ ..."}
```

## Test plan

- [x] `qmd search "term" --json` includes `line` field in each result
- [x] `qmd query "term" --json` includes `line` field in each result
- [x] `qmd search "term" --json --full` omits `line` (no snippet extraction when showing full body)
- [x] Existing CLI tests pass (`npx vitest run test/cli.test.ts` — 91/91)

Closes #505

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -1932,7 +1932,8 @@ function outputResults(results: OutputRow[], query: string, opts: OutputOptions)
     const output = filtered.map(row => {
       const docid = row.docid || (row.hash ? row.hash.slice(0, 6) : undefined);
       let body = opts.full ? row.body : undefined;
-      let snippet = !opts.full ? extractSnippet(row.body, query, 300, row.chunkPos, undefined, opts.intent).snippet : undefined;
+      const snippetInfo = !opts.full ? extractSnippet(row.body, query, 300, row.chunkPos, undefined, opts.intent) : undefined;
+      let snippet = snippetInfo?.snippet;
       if (opts.lineNumbers) {
         if (body) body = addLineNumbers(body);
         if (snippet) snippet = addLineNumbers(snippet);
@@ -1941,6 +1942,7 @@ function outputResults(results: OutputRow[], query: string, opts: OutputOptions)
         ...(docid && { docid: `#${docid}` }),
         score: Math.round(row.score * 100) / 100,
         file: toQmdPath(row.displayPath),
+        ...(snippetInfo && { line: snippetInfo.line }),
         title: row.title,
         ...(row.context && { context: row.context }),
         ...(body && { body }),


rymalia:
**CI note:** The `Bun (ubuntu-latest)` failure in `createStore throws without explicit path in test mode` is a pre-existing test isolation issue — the test doesn't call `_resetProductionModeForTesting()` before asserting, so leaked module state from prior Bun test files can suppress the throw. Unrelated to this PR's 3-line change in `outputResults()` JSON formatting.

**Re-running the job will likely pass — the failure depends on Bun's test file execution order.**

PR #535, opened 2026-04-08
Add a FreeBSD core runtime path

lin72h:
## Summary

This PR adds the first upstreamable Node.js path for running QMD core search on FreeBSD.

The implementation keeps the existing supported-platform behavior in place, uses a real `vec0.so` loadable extension for FreeBSD, and avoids depending on a downstream-vendored `node-llama-cpp` patch.

## What changed

- added a local sqlite-vec resolver that supports `QMD_SQLITE_VEC_PATH`
- updated the DB loading path to use a real loadable extension on FreeBSD
- made sqlite-vec diagnostics platform-aware instead of Homebrew-only
- made `node-llama-cpp` optional and conservative on FreeBSD so LLM commands fail fast instead of trying to auto-build through an unsupported path
- added FreeBSD docs and a smoke harness that distinguishes the immediate `--quick` core path from the later `--full` LLM follow-up

## Why this shape

FreeBSD is closer to the existing macOS system-library path than to the Linux prebuilt-binary path. The immediate upstream lane therefore prefers:

- a real system SQLite extension (`vec0.so`)
- source installs instead of a FreeBSD npm binary matrix
- graceful LLM unavailability until upstream `node-llama-cpp` support lands cleanly
- isolated platform handling instead of scattering `process.platform === "freebsd"` checks

## Validation

Validated on a real FreeBSD 15.0 amd64 host:

- clean `pnpm install --frozen-lockfile`
- `sqlite3` loading a FreeBSD-built `vec0.so`
- `qmd status`
- `test/freebsd-smoke.sh --quick`
- clean-room replay through the local review-kit and handoff scripts
- `qmd embed` failing with `node-llama-cpp is unavailable` instead of falling through to `spawn npm ENOENT`

## Known limits

- FreeBSD is Node-only for now
- this immediate upstream lane supports BM25 and sqlite-vec features now
- embeddings, query expansion, reranking, and full `test/freebsd-smoke.sh --full` remain gated on the adjacent upstream `node-llama-cpp` FreeBSD fix
- the initial packaging posture is source-install-first rather than a published FreeBSD binary matrix

Related upstream dependency: asg017/sqlite-vec#282

### DIFF
--- a/README.md
+++ b/README.md
@@ -475,11 +475,12 @@ The `query` command uses **Reciprocal Rank Fusion (RRF)** with position-aware bl
 ### System Requirements
 
 - **Node.js** >= 22
-- **Bun** >= 1.0.0
+- **Bun** >= 1.0.0 on supported Bun platforms
 - **macOS**: Homebrew SQLite (for extension support)
   ```sh
   brew install sqlite
   ```
+- **FreeBSD**: Node.js-only core path for now. BM25 and sqlite-vec search are supported immediately; embeddings, reranking, and query expansion follow the upstream node-llama-cpp FreeBSD fix. See [docs/FREEBSD.md](docs/FREEBSD.md) for the verified source-install path and current smoke/certification entrypoints.
 
 ### GGUF Models (via node-llama-cpp)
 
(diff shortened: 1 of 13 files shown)


tobi:

PR #538, opened 2026-04-08
fix(chunking): pair code fences by char and length

galligan:
## Pseudo-stack

```
main
└── #538  fix: code fence pairing               ← you are here (foundation)
        ├── #540  feat: list-aware chunking        (parallel with #541)
        └── #541  feat: XML tag regions             (parallel with #540)
```

**This PR: #538.** It's the foundation of the stack — no dependencies. The three PRs above it build on this fix. I'll rebase each of them onto the latest `main` as this one lands.

---

## Summary

Fixes code fence detection to follow CommonMark pairing rules so the chunker correctly recognizes fences longer than three backticks, tilde fences, and blocks that wrap shorter fences inside longer ones.

## The bug

`findCodeFences` had two problems:

1. The regex (`/
```/g`) only matched exactly three backticks and ignored tildes entirely.
2. Pairing was a naive toggle on every match, so any ` ``` ` flipped the open/close state regardless of length or character.

Consequences:

- A block opened with ` ```` ` (4 backticks) was never recognized as a fence, so the chunker could split inside it.
- Tilde fences (`~~~`) were invisible to the chunker.
- A stray ` ``` ` inside a longer fence would prematurely "close" it, leaving the real tail of the block looking like prose.

These are real content shapes — agent instruction files and tutorials routinely use ` ```` ` to wrap examples that contain ` ``` `.

## The fix

`findCodeFences` now tracks the opening fence's character and length. A later fence run closes the open fence only if:

- it uses the same character (`` ` `` or `~`), **and**
- its length is greater than or equal to the opener, **and**
- its line has no info string (nothing but whitespace after the run)

`BREAK_PATTERNS` now scans for runs of 3+ backticks or tildes as candidate break points.

## Scope & non-goals

- **Column-0 fences only.** Indented fences (≤3 spaces per CommonMark, or inside list items) are not detected. A comment on `findCodeFences` documents this.
- **No type or API changes.** `CodeFenceRegion`, `isInsideCodeFence`, `findBestCutoff`, and all call sites in `store.ts` are unchanged.
- **AST path untouched.** `src/ast.ts` doesn't process markdown, so nothing there needed to change.

## Same-length fences (CommonMark quirk)

You can't nest a 4-backtick block inside a 4-backtick block — the second ` ```` ` legitimately closes the first. To wrap a 4-bt block you need 5 backticks outside. This is CommonMark-spec behavior and matches every conforming parser; disambiguation would require either non-linear lookahead or content heuristics, both of which have worse failure modes than the spec's deterministic rule. Tests codify this so future changes don't accidentally "fix" it.

## Tests

12 new cases under `findCodeFences` in `test/store.test.ts`:

- 4-backtick wrapping 3-backtick
- Tilde fence basic
- Tilde not closed by stray backticks
- Longer fence not closed by shorter run
- Close line with info string is invalid
- 5/4/3 backtick nesting with bare inner fences
- 6-bt closes 5-bt (longer-than-opener close is valid)
- Same-length non-nesting (two regions, content outside)
- Mixed chars (backtick outer, tilde inner)
- Info strings on both outer and inner
- Tilde 5/4/3 nesting

All existing `findCodeFences` / `isInsideCodeFence` / `scanBreakPoints` tests continue to pass without modification. Full `test/store.test.ts` suite (203 tests) and `test/ast-chunking.test.ts` suite (12 tests) green locally.

## Test plan

- [x] `npx vitest run test/store.test.ts` passes locally (203/203)
- [x] `npx vitest run test/ast-chunking.test.ts` passes locally (12/12)
- [x] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Fixes
+
+- Code fence detection now follows CommonMark pairing rules. Fences
+  opened with 4 or more backticks (or tildes) are correctly recognized
+  and paired, so chunks no longer split inside nested code blocks that
+  wrap shorter fences. Tilde fences are now supported.
+
 ## [2.1.0] - 2026-04-05
 
 Code files now chunk at function and class boundaries via tree-sitter,
--- a/src/store.ts
+++ b/src/store.ts
@@ -101,7 +101,7 @@ export const BREAK_PATTERNS: [RegExp, number, string][] = [
   [/\n#{4}(?!#)/g, 70, 'h4'],      // #### but not #####
   [/\n#{5}(?!#)/g, 60, 'h5'],      // ##### but not ######
   [/\n#{6}(?!#)/g, 50, 'h6'],      // ######
-  [/\n```/g, 80, 'codeblock'],     // code block boundary (same as h3)
+  [/\n(?:`{3,}|~{3,})/g, 80, 'codeblock'],  // code block boundary (same as h3)
   [/\n(?:---|\*\*\*|___)\s*\n/g, 60, 'hr'],  // horizontal rule
   [/\n\n+/g, 20, 'blank'],         // paragraph boundary
   [/\n[-*]\s/g, 5, 'list'],        // unordered list item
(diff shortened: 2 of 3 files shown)


galligan:

PR #539, opened 2026-04-08
refactor(chunking): rename CodeFenceRegion to ProtectedRegion

galligan:
## Pseudo-stack

```
main
    └── #539  refactor: rename to ProtectedRegion   ← you are here
        ├── #540  feat: list-aware chunking            (parallel with #541)
        └── #541  feat: XML tag regions                 (parallel with #540)
```

**This PR: #539.** Stacked on #538. The diff below shows the fence pairing fix commits from #538 in addition to the rename — they'll collapse once #538 merges and I rebase. The two feature PRs above (#540 and #541) both depend on this rename.

---

## Summary

Pure rename, no behavior change. `CodeFenceRegion` becomes `ProtectedRegion` with an optional `kind` tag. This opens the seam for future chunking passes to contribute other kinds of protected regions without touching the core chunker contract.

## Why now

The fence fix (#538) is the first producer of protected regions. Two follow-up PRs are in flight that both rely on the rename:

- **#540** — list-aware chunking. Adds a scanner that emits weighted **break points** at list item boundaries (not protected regions — list items can be arbitrarily long and must remain splittable when they exceed chunk size).
- **#541** — XML tag regions. Adds a scanner that emits weighted **break points** at paired `<tag>…</tag>` boundaries for agent-prompt-style markdown. Also not protected regions — same reasoning: tag bodies can be large and must remain splittable.

Neither of those PRs introduces new `ProtectedRegion` producers right now — they contribute break points. The rename matters because the shared contract (`findBestCutoff` takes protected regions to filter out split candidates inside them) is now semantically general rather than code-fence-specific, and future passes that do need hard protection (e.g. tables, blockquotes) will have a natural seam to slot into.

Landing the rename on its own keeps the feature PRs focused on behavior rather than mechanical renames, and avoids a merge race between the two parallel feature branches.

## Changes

- `interface CodeFenceRegion` → `ProtectedRegion` with optional `kind?: string` (set to `'fence'` by `findCodeFences`)
- `isInsideCodeFence` → `isInsideProtectedRegion`
- `findBestCutoff` parameter: `codeFences` → `protectedRegions`
- `chunkDocumentWithBreakPoints` parameter: `codeFences` → `protectedRegions`
- Test-local variable names updated for consistency

`findCodeFences` keeps its name — it's now one producer of protected regions (and, for now, the only one). No external callers exist: `CodeFenceRegion` and `isInsideCodeFence` are not re-exported from `src/index.ts`, so the rename is fully contained within `src/store.ts` and its tests.

## Test plan

- [x] `npx vitest run test/store.test.ts` passes (203/203)
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Fixes
+
+- Code fence detection now follows CommonMark pairing rules. Fences
+  opened with 4 or more backticks (or tildes) are correctly recognized
+  and paired, so chunks no longer split inside nested code blocks that
+  wrap shorter fences. Tilde fences are now supported.
+
 ## [2.1.0] - 2026-04-05
 
 Code files now chunk at function and class boundaries via tree-sitter,
--- a/src/store.ts
+++ b/src/store.ts
@@ -80,12 +80,14 @@ export interface BreakPoint {
 }
 
 /**
- * A region where a code fence exists (between ``` markers).
- * We should never split inside a code fence.
+ * A region of the document that the chunker must not split inside.
+ * Code fences are the first producer; future passes may contribute
+ * other kinds (e.g. list items, XML tag regions) to the same shape.
  */
-export interface CodeFenceRegion {
-  start: number;  // position of opening ```
-  end: number;    // position of closing ``` (or document end if unclosed)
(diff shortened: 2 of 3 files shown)


galligan:

PR #553, opened 2026-04-10
feat(chunking): code fence pairing, list-aware + XML tag break points

galligan:
> This consolidates what was previously four stacked PRs (#538, #539, #540, #541) into a single PR for simpler review. The stacked approach created unnecessary friction for a cross-fork contribution. Full context in the [overview gist](https://gist.github.com/galligan/a8d3cada1e89633117fcc3f2733e8f8a). Each logical change is its own commit if you prefer to review commit-by-commit.

## Summary

Four chunking improvements, each in its own commit:

1. **Fix code fence pairing** to follow CommonMark rules (char + length matching, tilde support)
2. **Rename `CodeFenceRegion` to `ProtectedRegion`** with an optional `kind` tag (pure mechanical refactor)
3. **List-aware break point scanner** with depth-weighted scoring and nested sublist tracking
4. **XML tag break point scanner** for agent-prompt tags like `<example>`, `<instructions>`, `<thinking>`

Plus a test isolation fix for a pre-existing flaky test (`createStore throws without explicit path in test mode` on Bun ubuntu, same root cause as upstream commit 66e70c0).

## Splitting

These changes sit on clean seams and can be split into separate PRs if that's preferable. The commits are ordered so any prefix is self-contained:

- **Commits 1-2** (fence fix + rename) are a standalone bugfix with no feature additions
- **Commit 3** (list-aware) is independent of commit 4 (XML tags)
- **Commit 4** (XML tags) is independent of commit 3 (list-aware)

If you'd rather land these incrementally, I'm happy to split them back out.

## Commit 1: fix code fence pairing

`findCodeFences` had two problems: the regex only matched exactly three backticks (ignoring tildes and longer runs), and pairing was a naive toggle. A fence opened with ```` ```` ```` was never recognized. A stray ` ``` ` inside a longer fence prematurely closed it.

Now tracks the opening fence's character and length. A close candidate must use the same char, be at least as long, and carry no info string. `BREAK_PATTERNS` updated to scan for 3+ backticks or tildes.

**Scope:** column-0 fences only. Indented fences are not detected (documented).

**12 new tests** covering 4/5/6-backtick nesting, tilde fences, mixed chars, same-length non-nesting (CommonMark quirk), info-string validation.

## Commit 2: rename CodeFenceRegion to ProtectedRegion

Pure mechanical rename. `CodeFenceRegion` becomes `ProtectedRegion` with optional `kind?: string` (set to `'fence'` by `findCodeFences`). `isInsideCodeFence` becomes `isInsideProtectedRegion`. All parameter names updated. No behavior change. Not re-exported from `src/index.ts`, so fully contained.

## Commit 3: list-aware break point scanner

Replaces the two naive `BREAK_PATTERNS` entries (`list: 5`, `numlist: 5`) with `findListBreakPoints`, a stack-based scanner that tracks nested list frames and emits depth-weighted break points:

| Break point | Score |
|---|---|
| list-end (list to non-list transition) | 75 |
| Top-level item (depth 0) | 70 |
| Second-level item (depth 1) | 45 |
| Third-level and deeper (depth 2+) | 25 |

Handles: `-` and `*` unordered markers, `1.` and `1)` ordered markers, mixed marker characters at same indent (treated as one list), nested sublists, blank lines inside items, list-end detection.

Deliberately deferred: loose/tight distinction, lazy continuation, 4-space indented code blocks, tab indentation. Each documented in a block comment.

**16 new tests** including an end-to-end integration test through `chunkDocument` confirming long lists split at item boundaries.

## Commit 4: XML tag break point scanner

`findXmlTagBreakPoints` detects line-anchored paired XML tags and emits asymmetric break points:

| Break point | Score |
|---|---|
| tag-open | 30 (weak: splitting right before content orphans the opener) |
| tag-close | 75 (strong: splitting after a closed block is great) |

Tag name grammar: `[A-Za-z_][A-Za-z0-9_.:-]*` (XML Name production, custom elements, namespaced tags).

HTML5 element names are blocked via a case-insensitive blocklist in `src/html-elements.ts` so inline HTML (`<div>`, `<p>`, `<br>`) isn't confused for structural tags.

Key rules:
- Line-anchored only (tag must be on its own line, leading whitespace OK)
- Case-sensitive open/close matching (XML semantics)
- Stack-based nesting (same-name and different-name)
- Self-closing `<tag/>` produces nothing
- `<!-- -->`, `<!DOCTYPE>`, `<![CDATA[]]>`, `<?xml ?>` skipped
- Cross-tag interleaving is malformed (zero break points)
- Unclosed tags emit nothing
- Tags inside code fences are ignored (fence scan runs first)

**27 new tests** including fence precedence, HTML blocklist, case sensitivity, malformed input handling, and an integration test through `chunkDocument`.

## Regression analysis

The only pattern that used to score and no longer does is `-\t` (dash followed by literal tab as marker separator). The old regex `\n[-*]\s` matched it at score 5; the new scanner requires space-separated markers. Tab-indented list items were never detected by the old regex either (`\n\t- foo` was invisible). Not a regression in practice.

Everything the old code detected, the new code detects and scores higher. Previously undetected patterns (nested sublists, `1)` form, list-end transitions, XML tags, tilde fences, 4+ backtick fences) are now handled.

## Files changed

- **Modified:** `src/store.ts` (fence fix, rename, list scanner, XML scanner, integration)
- **Modified:** `test/store.test.ts` (55 new tests + test isolation fix)
- **Modified:** `CHANGELOG.md` (changelog entries under `[Unreleased]`)
- **New:** `src/html-elements.ts` (HTML5 element blocklist)

## Test plan

- [x] `npx vitest run test/store.test.ts` passes (246/246, was 203 + 43 new)
- [ ] CI green

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,17 @@
 
 ## [Unreleased]
 
+### Changes
+
+- List-aware chunking. A new scanner tracks nested list items with a
+  stack-based state machine and emits weighted break points: top-level
+  items score 70, second-level 45, third-level and deeper 25, and the
+  transition from a list back to prose scores 75. Previously the naive
+  `list: 5` and `numlist: 5` patterns produced break points too weak to
+  influence chunking. Long lists now split cleanly at item boundaries
+  instead of mid-item. Ordered `1)` form is newly supported, as is
+  proper detection of nested sublists (which the old regex missed).
+
 ### Fixes
 
 - Code fence detection now follows CommonMark pairing rules. Fences
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,17 @@
   instead of mid-item. Ordered `1)` form is newly supported, as is
   proper detection of nested sublists (which the old regex missed).
 
+- XML tag break points for agent-prompt markdown. The chunker now
+  recognizes line-anchored paired tags like `<example>…</example>`,
+  `<instructions>…</instructions>`, and `<thinking>…</thinking>` that
+  commonly appear in agent instruction files, and prefers to split at
+  the close of a tagged block (score 75) rather than mid-block. HTML5
+  element names (`<div>`, `<p>`, etc.) are blocked via a blocklist so
+  inline HTML in markdown is not confused for structural tags. Tags
+  inside code fences are ignored. Supports nested same-tag blocks,
+  namespaced tags (`<xsl:template>`), and custom elements
+  (`<my-widget>`).
+
 ### Fixes
 
 - Code fence detection now follows CommonMark pairing rules. Fences
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Fixes
+
+- Code fence detection now follows CommonMark pairing rules. Fences
+  opened with 4 or more backticks (or tildes) are correctly recognized
+  and paired, so chunks no longer split inside nested code blocks that
+  wrap shorter fences. Tilde fences are now supported.
+
 ## [2.1.0] - 2026-04-05
 
 Code files now chunk at function and class boundaries via tree-sitter,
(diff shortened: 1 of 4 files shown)


tobi:

PR #571, opened 2026-04-13
fix: embed and pull commands now respect models config and env vars

beorn:
## Summary

- `qmd embed` and `qmd pull` hardcoded `DEFAULT_*_MODEL_URI`, ignoring the `models:` section in `index.yml` and `QMD_EMBED_MODEL` / `QMD_RERANK_MODEL` / `QMD_GENERATE_MODEL` environment variables
- `qmd status` always displayed default model URIs even when custom models were configured
- Only `query` and `vsearch` honored custom models (via the `LlamaCpp` instance set up in `getStore()`)

## Fix

Adds a module-level `models` object initialized with defaults, resolved to configured values (config > env > default) during `getStore()` initialization. The `embed`, `pull`, and `status` commands read from this object instead of hardcoded defaults.

The catch path in `getStore()` also resolves models from env/defaults, so the POJO is never stale even if config loading fails.

## Motivation

Discovered while trying to use `Qwen3-Embedding-8B` via `models.embed` in `index.yml` — `qmd query` used the configured model correctly, but `qmd embed -f` always re-embedded with the default 300M model.

## Test plan

- [x] All 250 existing tests pass
- [x] Verified `qmd embed` prints configured model URI (was: always `embeddinggemma-300M`)
- [x] Verified `qmd status` shows configured model

Happy to add CLI-level tests for model config propagation if desired.

Relates to #494 (v2.1.0 fix only covered `LlamaCpp` init, not `embed`/`pull`/`status` code paths).

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -112,23 +112,36 @@ enableProductionMode();
 let store: ReturnType<typeof createStore> | null = null;
 let storeDbPathOverride: string | undefined;
 let currentIndexName = "index";
+/** Resolved model URIs: YAML config > env var > built-in default */
+const models = {
+  embed: DEFAULT_EMBED_MODEL_URI,
+  generate: DEFAULT_GENERATE_MODEL_URI,
+  rerank: DEFAULT_RERANK_MODEL_URI,
+};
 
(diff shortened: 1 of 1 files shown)


tobi:

PR #573, opened 2026-04-14
Fix Nix wrapper so `qmd` works on NixOS

surma:
`qmd` was failing on NixOS when loading `node-llama-cpp` under Bun.

### Problem

The flake package wrapper set:

```sh
LD_LIBRARY_PATH=${sqlite}/lib
```

which overwrote the runtime library path entirely. On NixOS, the `node-llama-cpp` native addon also needs the C++ runtime (`libstdc++.so.6`). Because that library was no longer visible, the prebuilt addon failed to load.

Once that happened, `node-llama-cpp` tried to fall back to building from source in its install location under `/nix/store`, which is read-only, so `qmd embed` failed with `NoBinaryFoundError`.

### Fix

- add a shared `runtimeLibraryPath`
- include:
  - `sqlite`
  - `stdenv.cc.cc.lib` on Linux
- use `makeWrapper --prefix` instead of `--set` for `LD_LIBRARY_PATH` / `DYLD_LIBRARY_PATH`

This preserves any existing runtime search path while ensuring the packaged binary can find both SQLite and the C++ runtime.

### Why this is needed

`qmd` is wrapped around Bun in the Nix package, and Bun loads the `node-llama-cpp` native module at runtime. On NixOS, that module does not work unless `libstdc++.so.6` is available in the wrapped environment.

### Verification

Built locally with:

```sh
nix build .#qmd
```

Verified that the rebuilt binary can run successfully on NixOS:

```sh
result/bin/qmd status
result/bin/qmd embed
```

`embed` now completes successfully instead of failing during llama initialization.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -92,6 +92,9 @@
           outputHashMode = "recursive";
         };
 
+        runtimeLibraryPath = pkgs.lib.makeLibraryPath ([ pkgs.sqlite ]
+          ++ pkgs.lib.optionals pkgs.stdenv.isLinux [ pkgs.stdenv.cc.cc.lib ]);
+
         qmd = pkgs.stdenv.mkDerivation {
           pname = "qmd";
           inherit version;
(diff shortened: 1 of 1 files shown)


daniel-fahey:
Hah, snap!

I was about to start work on a PR (with CUDA support) and then I realised `qmd` has already been packaged in https://github.com/numtide/llm-agents.nix with both CUDA and Vulkan support.

Try it out `nix run github:numtide/llm-agents.nix#qmd -- status`, maybe consider closing this PR if that works for you, maybe `qmd` should signpost to the derivation there written by @mulatta and @Mic92?

Mic92:
Happy to relicense all my changes

surma:
Oh hell yea, that’s way better

PR #574, opened 2026-04-14
fix(nixos): fix prebuilt binaries via LD_LIBRARY_PATH + skip source build on read-only install

### LINKED ISSUE #87: Emebeding fails on NixOS (node-llama-cpp)

jerome-benoit:
## Summary

On NixOS (and other immutable-root systems), `qmd embed` crashes with `EACCES` because node-llama-cpp tries to compile llama.cpp into its read-only `node_modules/` directory inside `/nix/store`. Even when source builds are skipped, the prebuilt binaries fail because they expect FHS paths like `/lib64/libc.so.6` which don't exist on NixOS.

Two fixes:
1. **`LD_LIBRARY_PATH` in wrapper** — the flake wrapper sets `LD_LIBRARY_PATH` to include Nix's glibc and libstdc++, covering what the prebuilt binaries need without modifying them
2. **skip source builds** when the install directory is read-only (`canWriteLlamaDir()`)

## Changes

- **flake.nix**: Add glibc and libstdc++ to `LD_LIBRARY_PATH` in the wrapper. No binary modification needed — `LD_LIBRARY_PATH` takes precedence over RUNPATH for `dlopen`.
- **src/llm.ts**: Add `canWriteLlamaDir()` that probes write access on node-llama-cpp's `llama/` directory. When unwritable, `ensureLlama()` passes `build: "never"` to `getLlama()`.
- **CHANGELOG.md**: Unreleased entry under Fixes.

Closes #87

## Testing

- [x] `nix build` succeeds
- [x] `./result/bin/qmd status` shows CPU backend loaded, no EACCES, no NoBinaryFoundError
- [x] `./result/bin/qmd embed` completes successfully

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Fixes
 
+- NixOS: fix `qmd embed` crash on immutable-root systems. Prebuilt
+  node-llama-cpp binaries expect FHS paths like `/lib64/libc.so.6`
+  which don't exist on NixOS. The flake wrapper sets `LD_LIBRARY_PATH`
+  to include Nix's glibc and libstdc++. When the install directory
+  is read-only, `getLlama()` uses `build: "never"` to skip source
+  builds that would fail with EACCES.
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
 - Fix: preserve original filename case in `handelize()`. The previous
   `.toLowerCase()` call made indexed paths unreachable on case-sensitive
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -194,6 +194,12 @@
 - MCP: make `qmd mcp --index <name>` use the selected index for both foreground and daemon HTTP servers instead of falling back to the default store. #343
 - Embedding: respect `QMD_EMBED_MODEL` consistently for vector indexing and vector-backed search, with default-model fallback when unset.
 - Config: use one home-directory resolver for YAML config and the default SQLite cache path, avoiding Windows CLI/MCP split-brain when `HOME` is unset.
+- NixOS: fix `qmd embed` crash on immutable-root systems. Prebuilt
+  node-llama-cpp binaries expect FHS paths like `/lib64/libc.so.6`
+  which don't exist on NixOS. The flake wrapper sets `LD_LIBRARY_PATH`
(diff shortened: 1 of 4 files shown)


tobi:
Superseded by #861, which rebases the still-valid NixOS fix (LD_LIBRARY_PATH + skip source build on a read-only llama dir) onto current main without the incidental flake.lock nixpkgs bump.

PR #577, opened 2026-04-15
fix(bin): rewrite launcher as cross-platform Node.js script (fixes Windows global install)

kaimys:
## Problem

`npm install -g @tobilu/qmd` on Windows installs successfully but produces a `qmd` command that fails in every shell (cmd, PowerShell, Git Bash).

Root cause: `bin/qmd` has a `#!/bin/sh` shebang. When npm generates its cmd/ps1/sh shims on Windows, it reads that shebang and routes them through `/bin/sh` — a path that doesn't exist on Windows:

```cmd
IF EXIST "%dp0%/bin/sh.exe" (
SET "_prog=%dp0%/bin/sh.exe"
) ELSE (
SET "_prog=/bin/sh"
```

So the shim dies before any qmd code runs.

## Fix

Rewrite `bin/qmd` as a Node.js script with a `#!/usr/bin/env node` shebang. npm then generates native shims that invoke `node` directly on every platform. The runtime-detection logic is preserved verbatim:

- `package-lock.json` present → `node` (npm install)
- `bun.lock` / `bun.lockb` present → `bun`
- no lockfile → `node`

Implementation notes:
- Node is launched via `process.execPath` (guaranteed to exist, skips PATH resolution, avoids the `child_process` `shell: true` deprecation warning).
- Bun is launched by name; on Windows this requires `shell: true` so `PATHEXT` resolves `bun.exe` / `bun.cmd`.
- Symlink resolution still happens via `fs.realpathSync`, matching the old shell `readlink` loop, so `npm link` keeps working.
- Signals from the parent are re-raised after the child exits, so Ctrl-C behaves the same as before.

## Testing

- Verified locally on Windows 11: `qmd --help` works from cmd, PowerShell, and Git Bash after `npm install -g`.
- `test/launcher-detection.test.sh` still documents the intended detection behavior; its `detect_runtime` helper mirrors the JS implementation. A platform-agnostic vitest version can follow in a separate PR (would also let Windows CI catch regressions — currently CI only runs Linux + macOS).

## Alternative considered

Pointing `bin.qmd` in `package.json` directly at `dist/cli/qmd.js` (which already has a Node shebang). That's simpler but drops the bun-vs-node lockfile detection, so I kept the launcher.

### DIFF
--- a/bin/qmd
+++ b/bin/qmd
@@ -1,32 +1,66 @@
-#!/bin/sh
-# Resolve symlinks so global installs (npm link / npm install -g) can find the
-# actual package directory instead of the global bin directory.
-SOURCE="$0"
-while [ -L "$SOURCE" ]; do
-  SOURCE_DIR="$(cd -P "$(dirname "$SOURCE")" && pwd)"
-  TARGET="$(readlink "$SOURCE")"
-  case "$TARGET" in
-    /*) SOURCE="$TARGET" ;;
-    *) SOURCE="$SOURCE_DIR/$TARGET" ;;
-  esac
-done
+#!/usr/bin/env node
+// Cross-platform launcher for qmd.
+//
+// Previously this was a POSIX shell script with `#!/bin/sh`, which meant npm
+// on Windows generated shims that tried to route through `/bin/sh` — a path
+// that doesn't exist on Windows, so `qmd` failed immediately after a global
+// install. Rewriting the launcher in Node.js lets npm generate native
+// cmd/ps1/sh shims that invoke `node` directly on every platform.
 
-# Detect the runtime used to install this package and use the matching one
-# to avoid native module ABI mismatches (e.g., better-sqlite3 compiled for bun vs node)
-DIR="$(cd -P "$(dirname "$SOURCE")/.." && pwd)"
+import { spawn } from "node:child_process";
+import { existsSync, realpathSync } from "node:fs";
+import { dirname, resolve } from "node:path";
+import { fileURLToPath } from "node:url";
 
-# Detect the package manager that installed dependencies by checking lockfiles.
-# $BUN_INSTALL is intentionally NOT checked — it only indicates that bun exists
-# on the system, not that it was used to install this package (see #361).
-#
-# package-lock.json takes priority: if it exists, npm installed the native
-# modules for Node.  The repo ships bun.lock, so without this check, source
-# builds that use npm would be incorrectly routed to bun, causing ABI
-# mismatches with better-sqlite3 / sqlite-vec (see #381).
-if [ -f "$DIR/package-lock.json" ]; then
-  exec node "$DIR/dist/cli/qmd.js" "$@"
-elif [ -f "$DIR/bun.lock" ] || [ -f "$DIR/bun.lockb" ]; then
-  exec bun "$DIR/dist/cli/qmd.js" "$@"
-else
-  exec node "$DIR/dist/cli/qmd.js" "$@"
-fi
+// Resolve symlinks so global installs (npm link / npm install -g) can find
+// the actual package directory instead of the global bin directory.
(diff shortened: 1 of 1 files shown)


rainhead:
This PR already contains a fix for a bug that is live on `main` today, and I hit it independently — worth flagging, because a later rewrite landed the cross-platform launcher **without** this detail.

## The line that matters

```js
// For Node, use process.execPath — guaranteed to exist and avoids relying on
// PATH resolution.
const runnerCmd = runnerName === "node" ? process.execPath : "bun";
```

`main` currently has, at `bin/qmd:162`:

```js
const runner = useSourceMode ? sourceRunner : (runnerName === "node" ? "node" : "bun");
```

and spawns that bare string at line 166. So the launcher re-resolves `node` from `PATH` at spawn time, which defeats the shebang no matter what it says.

## Why that breaks

`better-sqlite3` is a V8-ABI addon compiled at **install** time. It is bound to the specific Node install that built it — not to a version range. `engines: { node: ">=22.0.0" }` spans ABI 127 (22), 137 (24) and 147 (26), three mutually incompatible ABIs, so the declared range is a promise the compiled artifact cannot keep. (The bundled tree-sitter packages ship N-API prebuilds and are immune; `better-sqlite3` is not.)

Repro, on qmd 2.5.3 installed globally under Homebrew's Node 26 (ABI 147), in any repo whose version manager selects a different major — here nvm reading an `.nvmrc` pinning Node 24 (ABI 137):

```
was compiled against a different Node.js version using NODE_MODULE_VERSION 147.
This version of Node.js requires NODE_MODULE_VERSION 137.
```

Every command fails, and `qmd mcp` fails the same way, so the MCP server never completes a handshake — it shows up only as `Connection failed` / `Connection closed` in the client's logs, with the real cause buried in stderr.

Pinning the shebang alone is **not** sufficient — I tried that first. The trampoline still re-resolves `node` when it spawns the real CLI, so the child lands back on the wrong install. `process.execPath`, as this PR has it, is the correct fix: it names the running install rather than a version.

## Relationship to #319

[#319](https://github.com/tobi/qmd/issues/319) is the same mechanism and is closed. It was framed on the bun-vs-node axis (bun-compiled ABI 141 vs nvm's 127) and the fixes that closed it made the launcher *runtime*-aware — which runtime to pick, not which **install** of that runtime. Two Node majors on one machine falls straight through that, and any version manager (nvm, fnm, asdf, volta, mise) reproduces it, which makes it a common configuration rather than an exotic one.

## Ask

Whatever happens to the rest of this PR — it has been open since April and the launcher has moved on under it — the `process.execPath` line is a one-line, platform-neutral fix that is still missing from `main`. It would be worth cherry-picking even if the rest is superseded.

Happy to open that as a minimal standalone PR against current `main` if that is easier to land than reviving this one. I am running the equivalent change as a local patch to `bin/qmd` and it resolves the failure completely; verified with `qmd status`, a real `qmd search`, and `qmd mcp` starting cleanly, all under the mismatched Node.

One adjacent note: the failure surfaces as a raw `ERR_DLOPEN_FAILED`. An ABI-mismatch check with a "reinstall qmd under this Node" message would turn a fairly cryptic crash into a self-service fix, and would also cover [#672](https://github.com/tobi/qmd/issues/672).

*Investigated and written with Claude Code; every claim here — the `main` line numbers, the ABI values, and the repro — was verified against the installed package and the current `main` before posting.*

tobi:
The remaining live piece of this PR — spawn `process.execPath` instead of PATH `node`, so `better-sqlite3` loads against the Node that installed qmd — landed in #879.

The rest (Node shebang launcher, Windows shims) already shipped on main earlier. Closing this as superseded; the branch is too far behind to rebase.

tobi:
Superseded by the Node launcher already on main plus #879 (process.execPath).

PR #566, opened 2026-04-13
feat(update): add -c/--collection filter to qmd update

erlebach:
# fix(update): add -c/--collection filter to `qmd update`

## Summary

- `qmd update` now accepts `-c <name>` (or `--collection <name>`) to re-index a single collection instead of all collections
- Multiple `-c` flags are supported (e.g. `qmd update -c notes -c work`)
- The completion message now names the collection when only one was updated (e.g. `✓ 'notes' updated.` vs `✓ 3 collections updated.`)
- Fix: added `transaction` to the `Database` interface in `src/db.ts`, which was used in `store.ts` but missing from the type definition (caused a build failure under updated `better-sqlite3` type declarations)

## Behaviour

```sh
qmd update                    # unchanged — updates all collections
qmd update -c mynotes         # updates only the 'mynotes' collection
qmd update -c notes -c work   # updates two specific collections
qmd update -c nonexistent     # exits with error: "Collection 'nonexistent' not found."
```

## Files changed

- `src/cli/qmd.ts` — `updateCollections()` accepts optional `collectionFilter?: string[]`; filters and validates before iterating; updated completion message
- `src/db.ts` — added `transaction<T>` to the `Database` interface

PR #580, opened 2026-04-15
fix(embed): honor -c <collection> with scoped force-clear and shared-hash preservation

### LINKED ISSUE #558: qmd embed silently ignores -c/--collection flag (processes global pending instead of scoping)
## Summary

`qmd embed -c <name>` accepts the `-c` flag without error but does **not** limit work to the named collection. It processes the full global pending backlog regardless of which collection was named. Combined with `-f`, the behavior is additionally destructive across all collections (see below).

This is particularly painful when integrating qmd into another tool: a downstream project that registers per-project qmd collections and wants to rebuild embeddings for just its own collection ends up triggering hours of unrelated work on whatever else is pending globally — while holding any lock the calling tool took before invoking qmd.

## Environment

- qmd 2.1.0 (ac782bcd32)
- node v25.2.1
- macOS Darwin 25.3.0 arm64 (Apple Silicon)
- Installed via `npm install -g @tobilu/qmd`

### 1. `qmd embed` help does not list `-c`

```
qmd embed [-f]                - Generate/refresh vector embeddings
    --max-docs-per-batch <n>    - Cap docs loaded into memory per embedding batch
    --max-batch-mb <n>          - Cap UTF-8 MB loaded into memory per embedding batch
```

No `-c` / `--collection` flag is documented for the embed subcommand, yet the flag is accepted silently with no warning or error.

### 2. Non-destructive minimal reproduction

Create a tiny one-doc collection, capture global pending/vector counts, run `qmd embed -c <name>`, observe the deltas.

```bash
mkdir -p /tmp/qmd-scope-a
echo "alpha content one" > /tmp/qmd-scope-a/a1.md

qmd collection remove scope-a 2>/dev/null
qmd collection add /tmp/qmd-scope-a --name scope-a --mask "*.md"

qmd status | grep -E "Vectors|Pending"
# Vectors:  3498 embedded
# Pending:  56521 need embedding

qmd embed -c scope-a &
EMBED_PID=$!
sleep 60
kill -9 $EMBED_PID

qmd status | grep -E "Vectors|Pending"
# Vectors:  3658 embedded     (+160)
# Pending:  56498 need embedding (-23)
```

`scope-a` contains **one** small document with a single chunk. If `-c` were honored, embed would process that single chunk and exit (or at most add 1–2 vectors). Instead, in ~60 seconds:

- Vectors grew by **+160** (not +1)
- Pending dropped by **-23** (not -1)

Both deltas are incompatible with a collection-scoped run. The 160 new vectors and 23 cleared pending docs were from the global aradex backlog that had nothing to do with `scope-a`. The named collection's own work was a rounding error inside the global pass.

Same pattern observed with a wiki-ai collection holding 27 docs (CO pages + sources, ~150–200 chunks total expected): running `qmd embed -c wiki-ai` added 1,417 vectors in 6 minutes before I killed it, and was clearly trending toward a full global rebuild.

### 3. `-c` combined with `-f` is globally destructive

Running `qmd embed -c wiki-ai -f` on a system with an otherwise-healthy global index caused:

- Vectors: **70,980 → 0** (full global wipe)
- Pending: **44,156 → 57,138** (all previously-embedded docs flipped back to pending)

Only the wiki-ai collection (27 docs / ~200 chunks) was referenced in the invocation, but every vector across every collection (including the much larger aradex collection with ~57k docs) was invalidated and queued for re-embedding. This is consistent with the `-c` flag being dropped on the floor: the `-f` applied globally because the whole command was interpreted as an unscoped force-rebuild.

Recovery from a single `-c wiki-ai -f` invocation is multi-hour, since at the observed ~175 chunks/minute embedding rate on Apple Silicon, reprocessing 57k+ pending chunks is a ~5 hour job.

## Impact

I'm integrating qmd as the retrieval layer for a project-specific wiki tool ([ResearchDesk](https://github.com/rijul20/researchdesk)). The wiki layer:

1. Takes a lock on a per-wiki state file
2. Writes a set of new markdown pages
3. Calls `qmd update -c <wiki-name>` (works correctly — scope honored)
4. Calls `qmd embed -c <wiki-name>` (this issue — scope ignored)
5. Releases the lock

Because `qmd embed -c <wiki-name>` actually processes the full global backlog, step 4 could hold the lock for **hours** during any compile that coincided with a cron-driven qmd update elsewhere on the system. I had a compile hang for 38 minutes before I realized the embed was grinding through an unrelated 44k-chunk aradex backlog, and the only path forward was to kill the compile and patch my tool to skip `qmd embed` entirely.

Workaround on my side was to stop calling `qmd embed` from compile at all. New wiki chunks now fall back to BM25 until the next scheduled `qmd embed` run. That works, but it means my tool is effectively unable to produce fresh vector embeddings for any specific collection on demand — a capability the CLI appears to advertise via its acceptance of the `-c` flag.

## Suggested fixes (one of):

1. **Actually honor `-c` in `qmd embed`**: filter the pending set to docs belonging to the named collection(s) before processing. This matches the behavior of `qmd update -c <name>`, which is correctly scoped. I'd guess this is the intended design given that `-c` is accepted without error.
2. **Reject `-c` at parse time** in `qmd embed` with an actionable error like `qmd embed does not support --collection scoping; use 'qmd embed' globally or filter via --max-docs-per-batch`. Worse than option 1 but at least avoids the silent footgun.
3. **Make `-f` respect collection scope** independently: `qmd embed -c X -f` should only invalidate X's vectors, not the whole index. Today it is a one-line command that wipes hours of embedding work across unrelated collections.

Any one of these would have saved me a ~40-minute debugging session plus the 5-hour recovery tail from the accidental `-f`.

## Related

Possibly related to multi-collection flag handling issues already reported:
- #181 (BM25 search with multiple `-c` flags)
- #233 (multi-collection `-c` filtering)
- #83 (SQLiteError when using `-c` with vsearch)

The common thread: `-c` is parsed inconsistently across qmd subcommands. For `embed` specifically, it appears to be parsed and discarded.

lukeboyett:
Fixes #558.

## Summary

`qmd embed -c <collection>` accepts the flag but ignores it: the handler never
reads `cli.opts.collection`, so a scoped embed ends up embedding every pending
document in the index. This PR makes `-c` behave on `embed` the way it does on
`search` / `query` / #566's forthcoming `update -c`, and tightens the adjacent
`--force` and vec-table paths that scoped embedding exposes.

Five focused commits:

1. **`fix(embed): honor -c <collection> by filtering pending docs`** — thread
   `options.collection` through `generateEmbeddings` → `getPendingEmbeddingDocs`
   / `getHashesNeedingEmbedding` with an added `AND d.collection = ?` filter.
   CLI handler forwards `cli.opts.collection`.

2. **`fix(embed): scope --force clear to the requested collection`** — when
   called with `force: true` *and* `collection`, `clearAllEmbeddings` now
   deletes only rows whose hash belongs to a doc in that collection and prunes
   the matching `vectors_vec` hash_seq entries, instead of wiping every
   collection's vectors. (vec0 virtual tables don't reliably accept
   IN-subquery DELETEs, so we enumerate and delete per row.)

3. **`fix(embed): preserve shared hashes when force-clearing a collection`** —
   `content_vectors` is keyed globally by content hash; two docs in different
   collections with identical bodies share one row. The scoped clear now only
   removes hashes *owned exclusively* by active docs in the target collection,
   so shared vectors stay valid for sibling collections.

4. **`fix(embed): validate -c against configured collections`** — route
   `cli.opts.collection` through `resolveCollectionFilter(..., false)` so an
   unknown collection errors consistently with the rest of the CLI, instead
   of silently reporting "0 to embed." This matches the pattern #566 uses for
   `update -c`.

5. **`fix(embed): drop vectors_vec when scoped force empties content_vectors`**
   — if the scoped clear empties `content_vectors` entirely (e.g. the target
   is the only active collection), drop `vectors_vec` so the next
   `generateEmbeddings` run recreates it via `ensureVecTable` with the current
   model's dimensions, matching the unscoped branch's behavior.

## Test plan

- [x] `npm test` — passes with new regression coverage in `test/store.test.ts`:
  - scoped embed embeds only target collection, sibling untouched
  - `-f -c` preserves sibling collection vectors
  - shared-hash preservation across collections
  - unknown `-c` value errors cleanly
  - scoped force rebuilds `vectors_vec` when `content_vectors` is emptied

## Dependencies and related work

- **Complements #566** (`feat(update): add -c/--collection filter to qmd update`). Same `resolveCollectionFilter(..., false)` validation path, so the two should compose without conflict regardless of merge order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,9 @@
   thread a SQL filter through `getPendingEmbeddingDocs` and
   `getHashesNeedingEmbedding`. When combined with `--force`,
   `clearAllEmbeddings` is also scoped to that collection so sibling
-  collections' vectors are preserved.
+  collections' vectors are preserved. Because `content_vectors` is
+  keyed by a global content hash, hashes shared with active documents
+  in other collections are left in place rather than re-deleted.
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
 - Fix: preserve original filename case in `handelize()`. The previous
   `.toLowerCase()` call made indexed paths unreachable on case-sensitive
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,12 @@
 
 ### Fixes
 
+- Embedding: `qmd embed -c <collection>` now actually filters by the
+  requested collection. Previously the flag was accepted but ignored —
+  `embed -c alpha` would embed every unembedded document across every
+  collection. `EmbedOptions.collection` (SDK) and the CLI `-c` flag now
+  thread a SQL filter through `getPendingEmbeddingDocs` and
+  `getHashesNeedingEmbedding`.
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
 - Fix: preserve original filename case in `handelize()`. The previous
   `.toLowerCase()` call made indexed paths unreachable on case-sensitive
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,9 @@
   `embed -c alpha` would embed every unembedded document across every
   collection. `EmbedOptions.collection` (SDK) and the CLI `-c` flag now
   thread a SQL filter through `getPendingEmbeddingDocs` and
-  `getHashesNeedingEmbedding`.
+  `getHashesNeedingEmbedding`. When combined with `--force`,
+  `clearAllEmbeddings` is also scoped to that collection so sibling
+  collections' vectors are preserved.
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
 - Fix: preserve original filename case in `handelize()`. The previous
   `.toLowerCase()` call made indexed paths unreachable on case-sensitive
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -1676,7 +1676,7 @@ function parseChunkStrategy(value: unknown): ChunkStrategy | undefined {
 async function vectorIndex(
   model: string = DEFAULT_EMBED_MODEL_URI,
   force: boolean = false,
-  batchOptions?: { maxDocsPerBatch?: number; maxBatchBytes?: number; chunkStrategy?: ChunkStrategy },
+  batchOptions?: { maxDocsPerBatch?: number; maxBatchBytes?: number; chunkStrategy?: ChunkStrategy; collection?: string },
 ): Promise<void> {
   const storeInstance = getStore();
   const db = storeInstance.db;
@@ -1686,7 +1686,7 @@ async function vectorIndex(
   }
 
   // Check if there's work to do before starting
-  const hashesToEmbed = getHashesNeedingEmbedding(db);
+  const hashesToEmbed = getHashesNeedingEmbedding(db, batchOptions?.collection);
   if (hashesToEmbed === 0 && !force) {
     console.log(`${c.green}✓ All content hashes already have embeddings.${c.reset}`);
     closeDb();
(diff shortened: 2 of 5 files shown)


lukeboyett:
Closing — the collection-scoped embed this adds is now native upstream (getPendingEmbeddingDocs(db, collection?, model) via 5b9f472 / b5f156c). Thanks for the work landing there.

PR #581, opened 2026-04-16
perf(store): add opt-in TTL cache to getStatus for polling callers

lukeboyett:
## Summary

On large indexes (millions of rows, tens of GB of SQLite), the
`COUNT(DISTINCT d.hash)` scan inside `getHashesNeedingEmbedding`
dominates `getStatus()` cost. Long-lived processes that call
`store.getStatus()` repeatedly — any polling caller checking index
health or stats — pay the full scan cost on every call, even though
the result barely changes between polls.

This PR adds an **opt-in per-Store TTL cache** to `getStatus`.
Default behavior is unchanged; callers opt in by passing `{ ttlMs }`.

Four commits (the last three are Codex review-round fixes):

1. **`perf(store): add opt-in TTL cache to getStatus for polling callers`** —
   `Store.getStatus` signature gains an optional `{ ttlMs }`. A call
   with `ttlMs > 0` reuses the cached snapshot if its age is under the
   window; a call without options (or with `ttlMs: 0`) re-queries and
   clears the cache. Per-Store scope, so multiple stores don't
   cross-contaminate. No call sites changed — MCP, CLI, SDK keep their
   existing fresh-each-call semantics until they opt in.

2. **`fix(status): each caller's ttlMs is authoritative against data age`** —
   store `fetchedAt` on the cache entry rather than `expiresAt`, and
   compare `now - fetchedAt` against *this* caller's `ttlMs`. Otherwise
   a short-TTL caller could silently get a snapshot older than their
   window if a prior long-TTL caller seeded the cache.

3. **`fix(status): stamp fetchedAt after the query completes`** — stamp
   with a post-query `Date.now()`, not the pre-query one. Otherwise the
   snapshot is already aged by the query's duration — and the query's
   duration is exactly what motivates this cache on large indexes.

4. **`test(status-cache): reset production mode after the suite`** —
   `enableProductionMode()` flips a module-global flag; the suite
   enabled it in `beforeAll` but never restored, which would leak into
   later test files under `fileParallelism: false`. Restore in
   `afterAll`, matching existing store/MCP test patterns.

## Test plan

- [x] `npm test` — new `test/status-cache.test.ts` covers:
  - default calls re-query every time
  - `ttlMs > 0` skips re-query within the window
  - `ttlMs: 0` behaves identically to no options
  - a fresh (no-options) call invalidates the cache
  - mixed-TTL callers: short-TTL re-fetches despite prior long-TTL seed
  - `fetchedAt` stamped after the query, not before
  - `enableProductionMode` flag restored in `afterAll`

## Dependencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## [Unreleased]
 
+### Changes
+
+- `store.getStatus(options?)` now accepts an optional `ttlMs`. Default
+  behavior is unchanged (every call runs a fresh query). Long-lived
+  processes that poll status (MCP servers, HTTP wrappers, monitoring
+  loops on large indexes) can pass `{ ttlMs: 5000 }` to reuse a recent
+  snapshot and avoid re-running the `COUNT(DISTINCT d.hash)` scan on
+  every request. Cache scope is per-`Store` instance; a subsequent
+  call without `ttlMs` (or with `ttlMs: 0`) invalidates the cache.
+
 ### Fixes
 
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
--- a/src/db.ts
+++ b/src/db.ts
@@ -69,6 +69,7 @@ export interface Database {
   exec(sql: string): void;
   prepare(sql: string): Statement;
   loadExtension(path: string): void;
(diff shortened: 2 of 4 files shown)


lukeboyett:
Closing — superseded by the store/status surface redesign in b5f156c / da184e5 (getStatus now handles this). Thanks!

PR #582, opened 2026-04-16
feat(bin): opt-in daemon fast-path for qmd search via QMD_DAEMON_URL

lukeboyett:
## Summary

Upstream already exposes `POST /search` (and `/query` alias) on the MCP HTTP
daemon (`src/mcp/server.ts`). Spinning up the full Node CLI for every
`qmd search` call — tokenizer load, SQLite open, etc. — is wasted work when a
daemon is already running. This PR teaches `bin/qmd` an **opt-in fast-path**
that POSTs to the daemon when `QMD_DAEMON_URL` is set, and falls back to the
normal CLI on any error.

Zero changes to server-side code. No call sites changed. Everything is
opt-in via a single env var.

## Behavior

- **`QMD_DAEMON_URL` unset** → unchanged behavior; the shim spawns the full
  CLI as before.
- **`QMD_DAEMON_URL=http://127.0.0.1:7433`** + `qmd search "query"` → a
  1-second `/health` probe, then a 30-second `POST /search`. Output is
  buffered so a failed request can't leak partial bytes before fallback.
- **Any failure mode falls through silently to the normal CLI**: env unset,
  `curl` missing, health probe fails, non-2xx response, empty query, unknown
  flag, etc. The user never sees a fast-path-specific error.
- `vsearch` is **not** routed through the fast-path — its semantics can't be
  faithfully expressed through `/search` (dropped in commit `d5b01e9` after
  review).

## Commits

One feature commit plus nine hardening follow-ups from review rounds:

- **`feat(bin): opt-in daemon fast-path for search and vsearch`** — initial
  fast-path with `QMD_DAEMON_URL` gating, `/health` probe, POST to `/search`.
- **`fix(bin): preserve argv + accept -n + accumulate multi -c`** — keep the
  shim's argv intact on fallback; support `-n` (CLI's limit flag) and
  repeated `-c <collection>`.
- **`fix(bin): fall through on unknown flags instead of mangling the query`** —
  any flag we don't recognise defers to the full CLI rather than silently
  being interpreted as part of the query string.
- **`fix(bin): guard value-taking flags against missing arg before shift`** —
  avoid `shift` errors when `-c` or `-n` is the last argv token.
- **`fix(bin): match CLI default limit of 5 in fast-path`** — align with the
  full CLI's documented default when `-n` is absent.
- **`fix(bin): normalise -n to match parseInt(n, 10) semantics`** — three
  review rounds iterating on `-n 3abc`, `-n ""`, `-n -5` to behave the way
  Node's `parseInt(n, 10)` does (or cleanly refuse).
- **`fix(bin): buffer daemon response so partial output can't leak before
  fallback`** — capture stdout, only emit on 2xx; on error, discard buffer
  and fall through to the CLI with the user seeing one clean result.
- **`fix(bin): drop vsearch from fast-path — its semantics can't route through
  /search`** — vsearch uses vector-only ranking and takes parameters `/search`
  doesn't accept; better to leave it on the full CLI path than produce subtly
  different results.

## Test plan

- [x] `npm test` — new coverage:
  - `test/bin-qmd-daemon-fastpath.test.ts` (284 lines): fast-path hits `/search`
    on healthy daemon, falls back on every failure mode above, argv
    preservation across fallback, `-n` / `-c` parsing equivalence with CLI,
    buffered-response invariants.
  - `test/mcp.test.ts` (+66 lines): `/search` REST handler contract the
    fast-path relies on.

## Dependencies

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,15 +4,19 @@
 
 ### Changes
 
-- `bin/qmd` gains an opt-in fast-path for `qmd search` / `qmd vsearch`.
-  When `QMD_DAEMON_URL` is set and points at a running
-  `qmd mcp --http --daemon`, these subcommands POST to the daemon's
-  existing `/search` REST endpoint instead of paying the per-call Node +
-  better-sqlite3 + sqlite-vec bootstrap. On large indexes this turns a
-  ~500–800 ms cold start into ~50–100 ms. Opt-in preserves the default
-  formatted-text output for interactive users; daemon mode prints the
-  JSON response for scripts/agents. `--index <name>` and any non-curl
-  error silently fall through to the cold-start CLI.
+- `bin/qmd` gains an opt-in fast-path for `qmd search`. When
+  `QMD_DAEMON_URL` is set and points at a running
+  `qmd mcp --http --daemon`, `search` POSTs to the daemon's existing
+  `/search` REST endpoint instead of paying the per-call Node +
+  better-sqlite3 + sqlite-vec bootstrap. On large indexes this turns
+  a ~500–800 ms cold start into ~50–100 ms. Opt-in preserves the
+  default formatted-text output for interactive users; daemon mode
+  prints the JSON response for scripts/agents. Unrecognised flags
+  (`--json`, `--min-score`, …), `--index <name>`, non-plain-positive
+  `-n` values, and any curl error fall through to the cold-start CLI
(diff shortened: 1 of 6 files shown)


lukeboyett:
Closing — stale on our side, and shared model-serving is being handled upstream via #663 (qmd serve). Our deployment now uses a fork-specific always-on launcher, so this isn't something we're proposing upstream anymore. Thanks!

PR #590, opened 2026-04-21
fix: use case-insensitive matching for context path prefix lookup

JemicyChu:
## Problem

File-level context set via `qmd context add` is not returned in search results — all results fall back to root-level context only.

### Root Cause

`getContextForPath()` and `getContextForFile()` in `src/store.ts` use `startsWith` for path prefix matching, which is **case-sensitive**. The `documents` table stores paths as resolved from the filesystem (e.g. `l3/cacheservice.md`), while `store_collections.context` preserves the original casing from the `qmd context add` command (e.g. `L3/CacheService.md`). This mismatch causes `startsWith` to fail silently.

### Reproduction

1. `qmd collection add wiki/ --name wiki`
2. `qmd context add "qmd://wiki" "Root context"`
3. `qmd context add "qmd://wiki/L3/CacheService.md" "File-level context"`
4. `qmd search "cache"` — hits CacheService.md
5. Result only shows root context; file-level context is missing

## Fix

Apply `.toLowerCase()` to both the normalized path and context key prefix before comparison in both `getContextForPath()` and `getContextForFile()`.

## Tests

Added two test cases covering mixed-case path prefix matching:
- `getContextForFile` matches context with different case in path prefix
- `getContextForPath` matches context with different case

All 200 tests pass.

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -2504,12 +2504,12 @@ export function getContextForPath(db: Database, collectionName: string, path: st
 
   // Add all matching path contexts (from most general to most specific)
   if (coll.context) {
-    const normalizedPath = path.startsWith("/") ? path : `/${path}`;
+    const normalizedPath = (path.startsWith("/") ? path : `/${path}`).toLowerCase();
 
-    // Collect all matching prefixes
+    // Collect all matching prefixes (case-insensitive)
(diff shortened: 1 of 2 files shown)


tobi:

Issue #585, opened 2026-04-17
qmd update doesn't mark removed files as inactive — orphan documents persist


frankenmason:
## Environment

- qmd version: 2.0.1 (179d450)
- Install: npm via nvm (`@tobilu/qmd`)
- OS: Ubuntu 24.04
- Node: v24.13.1
- better-sqlite3 backend

## Issue

After running `qmd update`, documents whose underlying files have been removed
from the filesystem are **not** marked as `active = 0` nor deleted from the
`documents` table. They persist as "ghost entries" and are never cleaned up,
even on subsequent `qmd update` runs.

## Reproduction

1. Run `qmd update` on collections where some indexed files have been moved or deleted since the last index.
2. Observe `qmd status` — `Total: N files indexed` inflates with stale paths.
3. Inspect the index DB directly:

   ```sh
   python3 -c "
   import sqlite3, os
   con = sqlite3.connect(os.path.expanduser('~/.cache/qmd/index.sqlite'))
   cur = con.cursor()
   rows = cur.execute('''SELECT d.path, s.path as base
                         FROM documents d
                         JOIN store_collections s ON d.collection = s.name''').fetchall()
   total = len(rows)
   ghosts = sum(1 for p, base in rows
                if not os.path.exists(p if p.startswith('/') else os.path.join(base, p)))
   print(f'total={total} ghosts={ghosts}')
   "
   ```

## Actual Measurement (2026-04-17)

On a workspace with 819 indexed documents across 10 collections:

| Collection      | Indexed | File exists | Ghost |
|-----------------|---------|-------------|-------|
| docs            | 293     | 47          | 246   |
| agents          | 51      | 0           | 51    |
| memory          | 107     | 0           | 107   |
| obsidian-vault  | 323     | 8           | 315   |
| wiki            | 17      | 8           | 9     |
| skills          | 12      | 0           | 12    |
| (others)        | 16      | 0           | 16    |
| **Total**       | **819** | **63**      | **756 (92%)** |

**All 756 ghost rows have `active = 1`.** Running `qmd update` again does not
transition any to `active = 0`, and the output always shows `... 0 removed`.

## Expected Behavior

`qmd update` should either:

- **(a)** mark documents with missing files as `active = 0` (soft delete), or
- **(b)** `DELETE` them entirely from `documents`, `content`, `content_vectors`,
  and the FTS5 virtual table.

## Analysis (dist/store.js)

The schema already supports `active = 1` filtering in read queries
[...]

PR #592, opened 2026-04-21
fix(store): preserve inactive document references in cleanupOrphanedContent

costajohnt:
Refs #585

## Problem

`cleanupOrphanedContent` deletes `content` rows not referenced by any **active** document:

```sql
DELETE FROM content
WHERE hash NOT IN (SELECT DISTINCT hash FROM documents WHERE active = 1)
```

Combined with `documents.hash` `ON DELETE CASCADE` (schema at `src/store.ts:776`), this cascade-deletes rows that `reindexCollection` just marked `active=0` a few lines earlier. The soft-delete pattern is silently turned into a hard delete, which also breaks the reactivation path in `insertDocument` (`ON CONFLICT ... DO UPDATE SET ... active = 1`) because the row is gone when a file comes back.

### Reproduction (on `main`)

The new test at `test/store.test.ts` (`reindexCollection soft-deletes orphaned docs and cleanupOrphanedContent preserves their content rows (#585)`) fails on `main`:

- Index 5 files in a collection
- Delete 3, re-index
- `result.removed === 3` (correct, the deactivation loop fires)
- But `SELECT COUNT(*) FROM documents WHERE active = 0` returns **0** (should be 3) and `SELECT COUNT(*) FROM content` returns **2** (should be 5). The 3 tombstones got cascade-deleted.

## Fix

Drop `WHERE active = 1` from the subquery so that inactive documents still count as references. Explicit hard-deletion stays available via `deleteInactiveDocuments()` (which the cascade then handles correctly).

```diff
 export function cleanupOrphanedContent(db: Database): number {
   const result = db.prepare(`
     DELETE FROM content
-    WHERE hash NOT IN (SELECT DISTINCT hash FROM documents WHERE active = 1)
+    WHERE hash NOT IN (SELECT DISTINCT hash FROM documents)
   `).run();
   return result.changes;
 }
```

## Relationship to the original report

The reporter in #585 observes ghost rows with `active = 1` and output showing `0 removed`. That suggests their specific scenario may involve an additional path I couldn't reproduce locally (possibly a legacy path-format mismatch between stored rows and `handelize()`, or a dist/source divergence in the npm-published `v2.0.1`). This PR fixes one concrete, reproducible bug on the same code path, but may not be the complete picture for what they're seeing. Happy to investigate further if you have pointers.

## Follow-up worth noting (not in this PR)

`removeCollection` at `src/store.ts:2675` uses the same `WHERE active = 1` pattern for its inline content cleanup. The semantics are slightly different (it runs after a hard `DELETE FROM documents WHERE collection = ?`), but a cross-collection soft-delete could be affected similarly. Happy to file a separate issue or PR if useful.

## Test plan

- [x] New regression test `reindexCollection soft-deletes orphaned docs and cleanupOrphanedContent preserves their content rows (#585)` in `test/store.test.ts`. Fails on `main`, passes after this fix.
- [x] Full `test/store.test.ts` passes (197/199. The 2 `LlamaCpp Integration > rerank` failures are pre-existing on `main`, unrelated to this diff.)
- [x] `tsc --noEmit` shows no new errors (two pre-existing type errors on `main` are unchanged)

### DIFF
--- a/src/store.ts
+++ b/src/store.ts
@@ -1949,13 +1949,20 @@ export function deleteInactiveDocuments(db: Database): number {
 }
 
 /**
- * Remove orphaned content hashes that are not referenced by any active document.
+ * Remove orphaned content hashes that are not referenced by any document.
+ *
+ * Inactive (soft-deleted) documents still count as references — otherwise the
+ * `documents.hash ON DELETE CASCADE` foreign key would cascade-delete rows that
+ * `reindexCollection` has just marked `active = 0`, turning soft deletes into
+ * hard deletes and breaking the reactivation path in `insertDocument`.
+ * `deleteInactiveDocuments()` exists as the explicit hard-delete entry point.
+ *
  * Returns the number of orphaned content hashes deleted.
  */
 export function cleanupOrphanedContent(db: Database): number {
   const result = db.prepare(`
     DELETE FROM content
-    WHERE hash NOT IN (SELECT DISTINCT hash FROM documents WHERE active = 1)
+    WHERE hash NOT IN (SELECT DISTINCT hash FROM documents)
   `).run();
   return result.changes;
 }
--- a/test/store.test.ts
+++ b/test/store.test.ts
@@ -3059,6 +3059,11 @@ describe("Content-Addressable Storage", () => {
         .prepare(`SELECT COUNT(*) AS c FROM documents WHERE collection = ? AND active = 0`)
         .get(collectionName) as { c: number }).c,
     ).toBe(3);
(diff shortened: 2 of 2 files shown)


costajohnt:
Bumping this in case it slipped through. CI is green and happy to address any feedback when someone has a moment to review.

costajohnt:
Closing this since it's superseded. You landed the same one-line fix in 92aaded on May 9 with an equivalent regression test, so my PR doesn't add anything. Glad we ended up on the same approach.

Issue #483, opened 2026-03-29
qmd query/vsearch unstable after embedding on macOS Apple Silicon: grammar-stack crash + hangs


quqi1599:
Hi, thanks for maintaining qmd.

I hit a reproducible issue on macOS Apple Silicon where `qmd search` works, embeddings complete successfully, but `qmd vsearch` and `qmd query` are unstable:

- sometimes they hang indefinitely / for a very long time
- sometimes they fail during query expansion with:
  `Failed to accept token in sampler: Unexpected empty grammar stack after accepting piece: 0 (15)`

## Environment

- machine: Apple Silicon MacBook Air
- OS: macOS Darwin 25.3.0 arm64
- qmd CLI in use locally reports package version `1.0.0` from global install path
- qmd repo observed: `tobi/qmd`
- install path in my environment: `~/.bun/install/global/node_modules/qmd`
- backend uses node-llama-cpp underneath

## What works

- indexing works
- `qmd embed` completes successfully
- `qmd status` after embedding shows:
  - `Documents: 678 files indexed`
  - `Vectors: 2071 embedded`
- plain BM25 search works:
  - `qmd search 'session store lock'`
  - `qmd search 'OpenClaw cron vs heartbeat'`

### 1) `qmd query` can fail during expansion

Observed stderr:

```text
Expanding query...
error: Failed to accept token in sampler: Unexpected empty grammar stack after accepting piece: 0 (15)
```

I traced this to the structured query expansion path.

Relevant source locations in the installed package:

- `src/llm.ts` -> `expandQuery()` creates a grammar and calls `session.prompt(..., { grammar, ... })`
- `src/qmd.ts` -> `expandQueryStructured()` is used by both `vectorSearch()` and `querySearch()`

The grammar in `src/llm.ts` is currently:

```bnf
root ::= line+
line ::= type ": " content "\\n"
type ::= "lex" | "vec" | "hyde"
content ::= [^\\n]+
```

### 2) `qmd vsearch` can hang

Even with a short query like:

```bash
qmd --index recalltest vsearch 'session store lock' --files -n 3
```

it can run for >45s and time out in my wrapper.

### 3) `qmd query` appears to have a concurrency hazard in vector branches

In `src/qmd.ts`:

- `vectorSearch()` explicitly runs vector queries **sequentially** and contains a comment explaining that concurrent embed calls can hang the embedding context.
[...]

Issue #474, opened 2026-03-27
Reranking hangs indefinitely on macOS M4 Pro (Qwen3-Reranker, node-llama-cpp)


ilkerbbb:
### Environment
- QMD: v2.0.1
- macOS: 15.3.0 (arm64, M4 Pro)
- Node: v25.8.1

### Problem
`qmd query` hangs at "Reranking X chunks..." stage. Never completes.

### Reproduction
1. `qmd search "test"` → works (0.3s)
2. `qmd vsearch "test"` → works (14s)
3. `qmd query "test"` → hangs at "Reranking 40 chunks..."

### Diagnosis
- Fresh model download: same issue
- `GGML_METAL=0` (CPU-only): same issue
- Likely related to llama.cpp #17743 (Qwen3 Reranker returns no output on macOS)

### Workaround
Using `vsearch` mode (semantic) for now.

droganov:
~Same for me~

UPD, no it just took long:

Searching 6 queries...
Embedding 4 queries... (1.5s)
Reranking 40 chunks... (92.0s)

PR #600, opened 2026-04-23
Add Apple Silicon Metal guardrails

### LINKED ISSUE #368: QMD 2.0.1: SIGABRT crash on process exit due to Metal GPU cleanup assertion (macOS, Apple Silicon)
## Description

`qmd query` crashes with SIGABRT on process exit after completing the reranking step. The query results are produced correctly, but the process aborts during Metal GPU resource cleanup (`ggml_metal_device_free`). This breaks integrations like OpenClaw's `memory_search` which treat a non-zero exit code as a failed query and return empty results.

`qmd search` (BM25 only, no GPU models) works fine — no crash.

## Environment

- **QMD**: 2.0.1 (a0bd077aaf)
- **OS**: macOS 15.4.1 (Darwin 25.3.0 arm64)
- **Chip**: Apple M3
- **Node**: v25.4.0
- **Install method**: `npm install -g @tobilu/qmd@latest`
- **Upgraded from**: 1.0.7 (which worked fine with `query` mode)

## Steps to Reproduce

```bash
qmd query "test query" --json -n 2
```

Query expansion, search, embedding, and reranking all complete successfully. Results are printed. Then the process aborts on exit.

## Crash Trace

```
Reranking 40 chunks...
/Users/runner/work/node-llama-cpp/node-llama-cpp/llama/llama.cpp/ggml/src/ggml-metal/ggml-metal-device.m:612: GGML_ASSERT([rsets->data count] == 0) failed
WARNING: Using native backtrace. Set GGML_BACKTRACE_LLDB for more info.
0   libggml-base.dylib                  0x0000000105f3144c ggml_print_backtrace + 276
1   libggml-base.dylib                  0x0000000105f31638 ggml_abort + 156
2   libggml-metal.so                    0x0000000106271e6c ggml_metal_device_init + 0
3   libggml-metal.so                    0x00000001062727b0 ggml_metal_device_free + 24
4   libggml-metal.so                    0x0000000106273f44 _ZNSt3__16vectorINS_10unique_ptrI17ggml_metal_device25ggml_metal_device_deleterEENS_9allocatorIS4_EEED1B8ne200100Ev + 72
5   libsystem_c.dylib                   0x000000019d5f342c __cxa_finalize_ranges + 480
6   libsystem_c.dylib                   0x000000019d5f31ec exit + 44
7   libnode.141.dylib                   0x000000010846f438 ...
```

Process exits with SIGABRT.

## Impact

- **OpenClaw integration broken**: OpenClaw's `memory_search` tool calls `qmd query --json` under the hood. The SIGABRT exit causes OpenClaw to discard the results, returning empty arrays. This is the same symptom as openclaw/openclaw#28181.
- **Workaround**: Setting `searchMode: "search"` (BM25 only) in OpenClaw config avoids the crash but loses reranking quality.

## Analysis

This appears to be the same upstream ggml Metal cleanup issue referenced in:
- QMD commit `4131c82` (Dec 2025): *"llama.cpp Metal backend still crashes at process exit due to ggml-metal cleanup issues"*
- Issue #129 (same class of bug on Windows/Bun with Vulkan)

The crash happens during `__cxa_finalize_ranges` → `ggml_metal_device_free`, meaning the Metal device destructor fires during process exit cleanup and asserts that GPU resource sets aren't empty. The workaround in the Dec 2025 commit (not explicitly disposing llama resources, letting process exit handle it naturally) may have regressed in 2.0's refactored cleanup path.

## Possible Fix

Since the results are already produced and written to stdout before exit, one option would be to use `process._exit(0)` instead of the normal `process.exit(0)` to skip `__cxa_finalize_ranges` entirely — similar to how #129 was approached for Bun.

## Version that worked

QMD 1.0.7 did not exhibit this crash on the same machine with the same Node version.

### LINKED ISSUE #388: GGML_ASSERT crash in Metal reranking on Apple Silicon (M1) — qmd query SIGABRT on 2.0.x
**QMD version:** 2.0.1 (also reproduced on 2.0.0 — different failure: missing dist files)
**Working version:** 1.1.6
**OS:** macOS 15.3.0 (Darwin 25.3.0 arm64) — Apple M1
**Node.js:** v25.8.1
**Runtime:** bun 1.3.8 (global install via `bun install -g`)

## Description

`qmd query` crashes with a `GGML_ASSERT` failure during the reranking step on Apple Silicon (M1). The assertion fires in `ggml_metal_device_free` when the process exits after reranking completes.

## Steps to reproduce

```bash
bun install -g @tobilu/qmd@2.0.1
qmd query "any search term" -n 3
```

## Expected behavior

Query completes and returns ranked results.

## Actual behavior

Query expansion, searching, and embedding all succeed. The crash occurs during/after the reranking step:

```
Reranking 28 chunks...
ggml/src/ggml-metal/ggml-metal-device.m:612: GGML_ASSERT([rsets->data count] == 0) failed
Process exited with signal SIGABRT
```

Full backtrace shows the assertion fires in `ggml_metal_device_init` called from `ggml_metal_device_free`, suggesting a Metal resource cleanup issue on process exit.

## Notes

- `qmd search` (BM25 only, no reranking) works fine on 2.0.1
- `qmd update` and `qmd embed` work fine on 2.0.1
- `qmd query` works perfectly on 1.1.6 (same machine, same index)
- Likely related to the bundled `node-llama-cpp` version change between 1.1.6 and 2.0.x

luoyuctl:
## Summary
- apply Apple Silicon Metal workarounds before `node-llama-cpp` initializes
- disable Metal residency sets by default on macOS arm64 to avoid exit-time `GGML_ASSERT([rsets->data count] == 0)` crashes
- disable the Metal tensor probe on M5/A19-class chips to avoid shader probe failures before fallback
- fall back to retrieval order when reranking is unavailable instead of aborting the whole query
- add targeted tests and document the new opt-out in `CLAUDE.md`
- merge latest `upstream/main` and resolve the `src/llm.ts` / `test/llm.test.ts` conflicts

## Why
On Apple Silicon, `qmd query` currently fails in a few related ways:

1. Processes can abort on exit inside `ggml_metal_device_free` due to residency set cleanup assertions.
2. On M5-class devices, the Metal tensor probe can fail during shader compilation even though the rest of the Metal path is still usable.
3. When rerank context creation fails, `qmd query` exits non-zero even though retrieval already succeeded.

This patch keeps the GPU path enabled, applies the minimal Metal workarounds before llama init, and degrades reranking gracefully when the reranker backend is unavailable.

Recent reports indicate this remains relevant after #652, especially #674 and the Apple Silicon instability tracked in #474/#483.

## Verification
- `pnpm install --frozen-lockfile`
- `pnpm test:types`
- `pnpm build`
- `env -u QMD_EMBED_MODEL -u QMD_GENERATE_MODEL -u QMD_RERANK_MODEL pnpm exec vitest run test/llm.test.ts --reporter=verbose --testTimeout 60000 -t 'QMD_APPLE_METAL_WORKAROUNDS|Apple Silicon Metal workarounds|native llama stdout containment|QMD_LLAMA_GPU resolution|model name resolution'`
- `env -u QMD_EMBED_MODEL -u QMD_GENERATE_MODEL -u QMD_RERANK_MODEL pnpm exec vitest run test/store.test.ts --reporter=verbose --testTimeout 60000 -t 'Embedding Formatting|rerank falls back to retrieval order when reranker fails'`

## Notes
- `QMD_APPLE_METAL_WORKAROUNDS=off` opts out of the automatic Metal env patching if someone wants the old behavior.
- A broader local `test/llm.test.ts test/store.test.ts` run can still be sensitive to caller `QMD_*_MODEL` environment overrides and real model initialization. The focused verification above clears those env vars.

Fixes #368
Fixes #388
Refs #474
Refs #483
Refs #674


amzzzzzzz:
Fresh Apple Silicon data point that may help validate this path.

Environment:
- QMD: `2.5.1` (`3aae056b8d`)
- macOS: `26.5` arm64
- Chip: Apple M4
- Node: `v22.22.0`

What is stable here:
- `qmd doctor` is clean and detects GPU Metal/offloading.
- `qmd status` is clean.
- `qmd search ...` is stable, so OpenClaw is currently kept on `memory.qmd.searchMode=search`.

What still looks risky after #652:
- Plain `qmd query ...` on the Metal path can still return useful results and then abort on process cleanup with the same class of error:

```text
ggml-metal-device.m:612 GGML_ASSERT([rsets->data count] == 0)
```

I also tried one ad hoc `QMD_FORCE_CPU=1 qmd query ... -n 1` check today; it was still running past ~30s and I stopped it, which lines up with the broader query/rerank instability tracked in #474/#483.

So from this machine, #600 still looks relevant even after the merged JSON safe-exit fix in #652: the macOS arm64 residency-set guardrail plus graceful rerank fallback would be the exact class of mitigation I would want before switching OpenClaw memory back from `search` to richer `query` mode.

jaserNo1:
Hey @tobi 👋

Just wanted to add a real-world data point: this PR would unblock a production multi-agent setup running 21 agents on an M4 Mac Mini (macOS 26). We've been hitting the `GGML_ASSERT([rsets->data count] == 0)` crash consistently on every `qmd query` invocation, which forced us to fall back to `QMD_FORCE_CPU=1` as a workaround — losing the GPU acceleration that's the whole point of running on Apple Silicon.

The fix in this PR is exactly what's needed: keeping Metal enabled but patching the residency set cleanup path that triggers the abort. The graceful rerank fallback is a nice bonus on top.

Regarding the `pnpm build` TypeScript error in `src/store.ts` mentioned in the PR notes — happy to submit a follow-up fix for that if it's blocking the merge. It's a pre-existing issue unrelated to this patch, but understandable if it holds things up.

Would love to see this land in 2.5.3 🙏

luoyuctl:
Thanks for the fresh Apple Silicon reports. I updated this PR against the current `main` and resolved the merge conflicts in `src/llm.ts` and `test/llm.test.ts`.

The resolution preserves the newer lazy `node-llama-cpp` import/stdout containment work from `main`, while still applying the Apple Metal env guardrails before the native module initializes.

Local verification after the merge:

- `pnpm install --frozen-lockfile`
- `pnpm test:types`
- `pnpm build`
- focused `test/llm.test.ts` coverage for Apple Metal workarounds, GPU mode resolution, model resolution, and native stdout containment
- focused `test/store.test.ts` coverage for embedding formatting and rerank fallback

I also updated the PR body to reference the newer reports in #474, #483, and #674, since this still appears relevant after #652.

luoyuctl:
Correction to my previous validation note: I have now force-pushed `0d77094`, rebased onto the current `main`, and resolved the conflicts while preserving the newer lazy import/stdout containment changes and the Apple Metal guardrails. `git diff --check` and conflict-marker checks pass.

This checkout currently cannot complete the TypeScript/build checks because `node_modules` is missing `@modelcontextprotocol/server` and its stdio types; `pnpm install --frozen-lockfile` is also blocked locally by pnpm native-build-script policy. Please rely on CI or a clean install for the final build verification. The M4/M5 reports and the rerank fallback remain covered by the updated code/tests.

luoyuctl:
Follow-up: the dependency mismatch is now fixed as well. I refreshed `pnpm-lock.yaml` in `68388f3`, then verified with:

- `pnpm install --frozen-lockfile --ignore-scripts`
- `./node_modules/typescript/bin/tsc -p tsconfig.build.json --noEmit`
- `node scripts/build.mjs`

All three pass. Native runtime tests still need the platform bindings/clean CI environment.

PR #604, opened 2026-04-26
fix: lazy load node-llama-cpp for no-op embed

aceclaw826:
## Summary

Avoid importing `node-llama-cpp` until QMD actually needs a local LLM/model operation.

This keeps no-op embed/status-style paths lightweight. In particular, `store.embed()` already checks for pending documents before calling into the LLM, but the previous static import meant the native `node-llama-cpp` module was still loaded as soon as `llm.ts` was imported.

This matters for OpenClaw multi-agent gateways where boot/update cycles may call `qmd embed` frequently and most runs often have no missing embeddings.

## Changes

- Convert `node-llama-cpp` value imports to type-only imports.
- Add a memoized dynamic loader for `node-llama-cpp`.
- Load `getLlama`, `resolveModelFile`, `LlamaLogLevel`, and `LlamaChatSession` only from code paths that actually need them.
- Add a regression test proving a no-op embed does not import `node-llama-cpp`.

## Related

This addresses part of openclaw/openclaw#72144: avoiding heavy native/LLM initialization when all content hashes already have embeddings.

## Tests

Passed:

```bash
corepack pnpm exec vitest run test/llm-lazy-load.test.ts test/sdk.test.ts -t "lazy node-llama-cpp loading|store.embed rejects invalid batch limits|store.embed forwards batch limit options" --reporter=verbose
```

Also checked:

```bash
corepack pnpm exec tsc -p tsconfig.build.json --noEmit
```

That currently fails on main with an unrelated existing error:

```text
```

### DIFF
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -4,19 +4,24 @@
  * Provides embeddings, text generation, and reranking using local GGUF models.
  */
 
-import {
-  getLlama,
-  resolveModelFile,
-  LlamaChatSession,
-  LlamaLogLevel,
-  type Llama,
-  type LlamaModel,
-  type LlamaEmbeddingContext,
-  type Token as LlamaToken,
+import type {
+  Llama,
+  LlamaModel,
+  LlamaEmbeddingContext,
+  Token as LlamaToken,
 } from "node-llama-cpp";
 import { homedir } from "os";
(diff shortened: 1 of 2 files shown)


tobi:

PR #608, opened 2026-04-27
Daemon-aware CLI fast-path: ~4× speedup for qmd query

ofa1:
# Daemon-aware CLI fast-path: 4× speedup for `qmd query`

> Filed as draft for early feedback. Happy to redirect / split / reshape based on what you'd prefer here.

## Why

`qmd query` from the CLI cold-loads the embed model, the query-expansion model, and the reranker on every invocation. That's ~10–17s on my M3 Pro (1069 docs across pi-memory + obsidian + a couple wikis), even when `qmd mcp --http --daemon` is sitting right there with everything already in VRAM.

I measured the breakdown:

| Path | Cold | Warm |
|---|---|---|
| `qmd query` in-process | 10–17s | 0.75s |
| `POST /query` against the daemon | 3.4s | 30ms |

The reranker dominates the in-process tail — ~10–12s of "load + score 40 chunks." The daemon already has it warm. The CLI just doesn't talk to the daemon today.

## What this PR does

When a daemon is running and its index matches the CLI's target DB, `qmd query` POSTs to a new `POST /v1/search` endpoint instead of cold-loading models. Falls back to the existing in-process path silently on any failure.

Live before/after on this branch with the daemon running:

| Query (uncached) | `--no-daemon` | default (daemon route) |
|---|---|---|
| `"machine learning weekly reflection concept"` | **13.2s** | **3.3s** |

## Architecture (8 commits, atomic)

```
326d33e test(cli): use structured lex: queries in parity tests for CI safety
0d12eee test(daemon): cover QMD_NO_DAEMON and malformed-response paths
2274389 docs: document daemon-aware CLI fast-path
9f989f1 test(cli): daemon-vs-direct output parity and dbPath safety
aec6df2 feat(cli): --no-daemon flag, QMD_DEBUG, daemon URL in status
3b52622 feat(cli): route qmd query through daemon when DB paths match
daa4e18 feat(daemon): HTTP client for /v1/search with silent fallback
94d6249 feat(mcp): add POST /v1/search for rich HybridQueryResult responses
43fd159 feat(daemon): discovery + /health dbPath + mcp.port persistence
```

1. **`src/daemon.ts`** (new, 242 lines) — `discoverDaemon()` + `searchViaDaemon()`. Reads `~/.cache/qmd/mcp.pid` + a new `mcp.port` file, probes `/health` with 300ms timeout, never throws. Returns `null` on any failure so the caller can fall through.
2. **`POST /v1/search` on the daemon** — returns full `HybridQueryResult[]` (not the lossy snippet-only `/query`). Calls `hybridQuery` / `structuredSearch` directly, not `store.search()`, so `candidateLimit` is actually forwarded (the SDK's `SearchOptions` interface drops it on the floor — happy to fix that separately).
3. **`/health` extended** with `dbPath` and `version`. The CLI uses `dbPath` to refuse routing when the daemon is serving a different index. Old daemons (pre-this-PR) don't include `dbPath` and are skipped, so they keep working unchanged.
4. **`mcp.port` persisted** at `~/.cache/qmd/mcp.port` after `listen()` resolves. Cleaned up on `qmd mcp stop` and on graceful shutdown. Falls back to 8181 if missing.
5. **`querySearch()` rewired** in the CLI to try the daemon first when the DB paths match (compared via `realpathSync` so symlinks don't cause false negatives). `qmd search` (BM25) and `qmd vsearch` (vec-only) intentionally stay in-process — `search` is already 0.2s, and `vsearch`'s pipeline is genuinely different from `hybridQuery`, so routing it would violate output parity. A future `/v1/vsearch` could unlock that.
6. **Escape hatches:**
   - `--no-daemon` per-invocation
   - `QMD_NO_DAEMON=1` global
   - `QMD_DEBUG=1` to log routing decisions to stderr (`Using daemon at http://localhost:8181 (PID …)` or `dbPath mismatch (…)`)
7. **`qmd status`** now shows the daemon URL: `MCP:   running (PID 66503) at http://localhost:8181`.
8. **No new runtime deps.** Built on `node:http`, `fetch`, `AbortController`.

## Output parity invariant

The PR's main correctness guarantee: daemon-mode and in-process mode must produce byte-identical output. Tested directly in `test/cli.test.ts`:

```ts
test("qmd query --json output matches with and without daemon", async () => {
  // Spawn real daemon, run CLI both ways with QMD_DEBUG=1, assert:
  // - daemon leg's stderr DOES contain "Using daemon" (so test can't pass trivially)
  // - --no-daemon leg's stderr does NOT
  // - docid arrays equal
  // - file arrays equal
});
```

Also tests the dbPath safety gate (`daemon with mismatched dbPath falls back to in-process`) and the `--no-daemon` flag.

## Tests

- 27 new tests across `test/daemon.test.ts`, `test/cli.test.ts`, `test/mcp.test.ts`.
- 817 total tests, 0 failing locally.
- 739 pass / 78 skipped under `CI=true` (existing `.skipIf(CI)` pattern for LLM-using tests; the new parity tests use `lex:` structured queries to skip the LLM expander, so they run in CI too).

`./node_modules/.bin/tsc -p tsconfig.build.json --noEmit` is clean.

## Backwards compatibility

- `POST /query` and the MCP protocol surface are untouched.
- `GET /health` only adds keys, doesn't remove any.
- An installed CLI from this PR talking to an older 2.0.x daemon: the older daemon's `/health` doesn't include `dbPath`, so `discoverDaemon()` returns null and the CLI falls back to in-process. Verified live against my pre-PR daemon — 0 functional regressions.
- `qmd status` line format gains a suffix: `MCP:   running (PID N) at http://localhost:P`. If anyone parses this line with `grep`, the `running` keyword is unchanged.

## One outside-Task type fix

`src/db.ts` gained one line: `transaction<T extends (...args: any[]) => any>(fn: T): T;` on the `Database` interface. The method is already used at `src/store.ts:2142` but the type was missing — `tsc --noEmit` was failing on it without this. Pulled into the same commit as Task 1 since it was needed to compile.

## Things I deliberately didn't do (happy to add if you want)

- **`POST /v1/vsearch`** so `qmd vsearch` can also use the warm embed model. Would need a thin endpoint that mirrors `vectorSearchQuery`. Felt out of scope for "make `qmd query` fast."
- **Add `candidateLimit` to `SDK.SearchOptions`.** I worked around it by calling `hybridQuery` / `structuredSearch` directly from the daemon handler. The SDK gap is a separate concern.
- **Skill / docs update for the new performance numbers.** I added a `## Making qmd query Faster` section to `skills/qmd/SKILL.md`. Happy to expand.

## Open questions for you

1. Is `/v1/search` versioning the right choice, or do you want this folded into `POST /query` with a `format: "rich"` flag?
2. The `dbPath` match uses a strict `realpathSync` comparison. Should it instead compare resolved DB **dirname + filename hash**, allowing for renamed-but-equivalent paths? I lean toward strict; figured I'd ask.
3. Anything you'd want to see benchmarked / measured before un-drafting?

Plan-doc with full design rationale (gitignored locally so not in this branch) lives in my session notes; happy to paste it here as a comment if useful.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 ## [Unreleased]
 
+### Changes
+
+- CLI fast-path: `qmd query` now routes through the MCP HTTP daemon
+  when one is running (`qmd mcp --http --daemon`) and the daemon's
+  index matches the CLI's target DB. Cold-start `qmd query` drops
+  from ~10-17s to ~2-4s by reusing the daemon's warm embedding,
+  expansion, and reranker models. Falls back transparently to the
+  in-process path when no daemon is reachable or when the DB differs
+  from the daemon's index. Disable with `--no-daemon` or
+  `QMD_NO_DAEMON=1`; enable stderr trace with `QMD_DEBUG=1`. `qmd
+  search` (BM25) and `qmd vsearch` (vector-only) remain in-process
+  unchanged. Adds a new versioned endpoint `POST /v1/search` that
+  returns full `HybridQueryResult[]` (existing `POST /query` is
+  unchanged) and extends `GET /health` with `dbPath` and `version`.
+  Daemon now persists `~/.cache/qmd/mcp.port` alongside `mcp.pid` so
+  the CLI can discover custom ports; cleaned up on `qmd mcp stop`.
+
 ### Fixes
 
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
--- a/README.md
+++ b/README.md
@@ -135,6 +135,27 @@ LLM models stay loaded in VRAM across requests. Embedding/reranking contexts are
 
 Point any MCP client at `http://localhost:8181/mcp` to connect.
 
+##### Daemon Fast-Path for the CLI
+
+When the HTTP daemon is running, `qmd query` automatically routes its
+search pipeline through the daemon instead of cold-loading the
+embedding, expansion, and reranker models for each invocation. On a warm
+daemon, `qmd query` drops from ~10-17s to ~2-4s (cold) or sub-second (warm
+rerank cache). `qmd search` (BM25) and `qmd vsearch` (vector-only)
+remain in-process — BM25 is already fast, and vsearch's pipeline is
+semantically different from the daemon's hybrid path.
+
+No configuration needed: the CLI discovers the daemon via
+`~/.cache/qmd/mcp.pid` + `~/.cache/qmd/mcp.port`. If the daemon is
+unreachable or the CLI is using a non-default index (`--index` or
+`INDEX_PATH`), it silently falls back to the existing in-process path.
+
+Escape hatches:
+- `--no-daemon` on a single invocation
+- `QMD_NO_DAEMON=1` environment variable (global)
+- `QMD_DEBUG=1` to log daemon discovery decisions to stderr
+- `qmd status` shows the daemon URL when running
+
 ### SDK / Library Usage
 
 Use QMD as a library in your own Node.js or Bun applications.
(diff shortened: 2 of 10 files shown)


tobi:

Issue #454, opened 2026-03-22
CJK query expansion produces English-only terms, degrading recall for Chinese queries


Seraphitah:
## Summary

When using `qmd query` with Chinese (CJK) input, the built-in query expansion model (`qmd-query-expansion-1.7B`) expands Chinese queries into English terms, which significantly degrades retrieval quality for Chinese-language document collections.

## Steps to Reproduce

1. Index a collection of Chinese Markdown files
2. Run a Chinese query:
   ```bash
   qmd query "Telegram 配置信息"
   ```
3. Observe that the expanded query contains English terms like "access sara's telegram account" instead of preserving the original Chinese semantics

## Expected Behavior

Query expansion should either:
- Preserve the original language of the query (expand Chinese → Chinese)
- Or provide a config option to disable expansion (e.g., `expandQuery: false` in `index.yml`)

## Current Behavior

The 1.7B expansion model appears to be trained primarily on English data. Chinese queries get "translated" into English during expansion, causing the vector search to match against English-like embeddings rather than the actual Chinese content.

## Workaround

Using typed query syntax (`lex:` / `vec:`) bypasses implicit expansion, but this is not available when QMD is called programmatically via MCP or integrated tools (e.g., OpenClaw `memory_search`).

## Suggestion

One or more of:
1. Add a `expandQuery: true|false` option in `index.yml` or per-collection config
2. Add a `--no-expand` flag to `qmd query`
3. Detect input language and skip expansion for non-English queries
4. Train/ship a multilingual expansion model

## Environment

- qmd version: latest (installed via bun)
- Embedding model: bge-m3 (GGUF)
- OS: macOS (arm64)
- Document language: Chinese (Simplified)

kechol:
+1. I ran into this on a Japanese corpus too, and suggestion No.2 (`--no-expand`) matched what I was looking for.

I drafted a small PR (#622) along those lines in case helpful, but happy to defer if a different approach from your list is preferred.

botAGI:
Can confirm the root cause: the expansion model's training data is 100% English, so any non-English query gets "translated" into English boilerplate (`...is an important concept... in software development`) and the output drifts off the format contract.

[...]

PR #622, opened 2026-05-04
feat: add --no-expand flag to qmd query

kechol:
#454

## Summary

Add a `--no-expand` flag to `qmd query` that skips LLM query expansion and searches the original query verbatim.

## Motivation

The built-in query expansion model (`qmd-query-expansion-1.7B`) is trained primarily on English. Given a non-English query — Chinese, Japanese, Korean — it translates the query into English before generating lex/vec/hyde variants. The vector search downstream then runs against English-translated text, and on a non-English corpus recall drops sharply.

A typed-query path (`{ type: 'lex', query: ... }` via SDK or MCP) bypasses expansion already, but for the simple ergonomic `qmd query "..."` CLI path there was no escape hatch. This PR adds one.

## Changes

- `qmd query --no-expand` skips expansion entirely and searches the original query verbatim. Mirrors `--no-rerank`.
- SDK: `search({ query, expand: false })` — analogous to the existing `rerank: false`.
- `HybridQueryOptions.skipExpansion` plumbs through to `hybridQuery()`. When set, `expandQuery()` is not called and the lex/vec/hyde fan-out is empty — only the original query runs against BM25 and the vector index.
- CLI hook output emits a single `Expansion disabled (--no-expand)` line on stderr so users see the flag took effect, instead of the misleading "Expanding query..." line.
- MCP `query` tool already takes pre-expanded `searches`, so expansion is structurally absent there. No MCP changes needed.

## Usage

```sh
# Skip expansion for non-English queries
qmd query "テレグラム 設定" --no-expand

# Same effect via SDK
const results = await store.search({ query: "テレグラム 設定", expand: false })
```

## Notes

- Default behavior is unchanged when `--no-expand` is absent (or `expand` is unset / `true`).
- Strong-BM25-signal short-circuit still works on the default path; `--no-expand` just makes the bypass unconditional.
- Pairs naturally with `--no-rerank` for the lowest-latency CPU path: `qmd query "..." --no-expand --no-rerank` runs only BM25 + vector + RRF, no LLM calls.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## [Unreleased]
 
+### Changes
+
+- `--no-expand` flag on `qmd query` skips LLM query expansion and
+  searches the original query verbatim. Mirrors the existing
+  `--no-rerank` flag and is also exposed as `expand: false` on the
+  SDK `search()` method. Useful for non-English corpora where the
+  expansion model translates the query into English and degrades
+  vector recall, and as a faster path when you trust your query
+  wording.
+
 ### Fixes
 
 - GPU: respect explicit `QMD_LLAMA_GPU=metal|vulkan|cuda` backend overrides instead of always using auto GPU selection. #529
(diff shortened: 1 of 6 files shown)


tobi:

PR #517, opened 2026-04-06
Add OpenAI-compatible remote embedding and reranking

jhsmith409:
## Summary

- Adds `RemoteLLM` class that calls OpenAI-compatible HTTP endpoints for **embedding (`POST /v1/embeddings`)**, **reranking (`POST /v1/rerank`)**, and **query expansion (`POST /v1/chat/completions`)**, with per-endpoint circuit breakers, dimension validation, batch splitting, auth headers, and configurable timeouts
- Adds `HybridLLM` compositor that routes embed/rerank/expand to a remote server when configured, falling back to local `LlamaCpp` for any path that isn't (tokenization always stays local)
- Generalizes the `LLM` interface with `embedBatch` and `embedModelName`, and updates the singleton/session management to accept any `LLM` implementation (backward-compatible)
- Configured via env vars (`QMD_EMBED_API_URL` / `QMD_EMBED_API_MODEL`, `QMD_RERANK_API_URL` / `QMD_RERANK_API_MODEL`, `QMD_EXPAND_API_URL` / `QMD_EXPAND_API_MODEL` / `QMD_EXPAND_API_KEY`) or YAML fields under `models:` (`embed_api_*`, `rerank_api_*`, `expand_api_*`)
- Skips nomic/Qwen3 text formatting prefixes for remote models (they handle their own prompt formatting)
- Zero new dependencies — uses Node.js built-in `fetch()`

## Motivation

Allows offloading embedding, reranking, **and query expansion** to a remote OpenAI-compatible server (e.g. vLLM, Ollama, LM Studio) while keeping QMD's local-first defaults intact. Useful when:
- The indexing machine doesn't have a GPU
- You want larger/better embedding or reranker models than what fits in local VRAM
- You already run a shared GPU inference server and want QMD to use it for the LLM-bound steps

Each remote endpoint is independently configurable — operators can mix-and-match (e.g. remote embed + remote rerank + local expand, or all-remote, or any subset). Unconfigured paths fall back to the existing local `LlamaCpp` path with no behavior change.

Related: #489, #427, #446, #511

## What's now remote-capable

| Path | Remote endpoint | Env vars | Local fallback |
|---|---|---|---|
| Embedding | `POST /v1/embeddings` | `QMD_EMBED_API_URL` + `QMD_EMBED_API_MODEL` | `LlamaCpp.embed()` |
| Batch embedding | `POST /v1/embeddings` (multi-input) | (same as embedding) | `LlamaCpp.embedBatch()` |
[...]

PR #629, opened 2026-05-05
Add remote embedding, reranking, and query expansion support

georgelichen:
This PR ports the remote embedding / reranking work from PR #517 onto the current upstream `main`, and includes the follow-up fixes needed to make it usable in the current tree.

## What is included
- add OpenAI-compatible remote embedding support
- add OpenAI-compatible remote reranking support
- add remote query expansion support via chat completions
- fix drift against current `main` (build/model label integration fixes)
- avoid initializing local `node-llama-cpp` during `qmd embed` when remote embedding is configured

## Why this PR exists
PR #517 was based on an older branch state. This branch rebases the feature set onto the current upstream `main` and resolves the integration drift, so the remote LLM path can be reviewed against today's tree.

## Verification
- `npx tsc -p tsconfig.build.json`
- `npx vitest run --reporter=verbose test/remote-llm.test.ts test/remote-llm-integration.test.ts`
- `npx vitest run test/store.test.ts -t "generateEmbeddings" --reporter=verbose`
- `npx vitest run test/store.test.ts -t "Token chunking guardrails" --reporter=verbose`

## Notes
- Query expansion uses `expand_api_*` when configured; otherwise normal `qmd query "..."` still falls back to local expansion.
- Structured queries (`intent:/lex:/vec:/hyde:`) skip auto expansion entirely.


Kaspre:
Running this PR (cherry-picked onto v2.5.1) in production. Endorsing.

Two minor rebase notes for anyone else picking this up against current main:

1. `vectorIndex()` needs `model = getDefaultLLM().embedModelName` set explicitly — that line was dropped during conflict resolution on our end.
2. One missed call-site rename: `getDefaultLlamaCpp` at `cli/qmd.ts:3716`.

Three small follow-ups we ended up needing in our local fork that you may want to fold in:

**A) `RemoteLLM.rerank` sigmoid normalization (~3 lines).** For rerankers exposed via llama.cpp's `/v1/rerank` (notably `bge-reranker-v2-m3`) and several other cross-encoder backends, `relevance_score` is log-odds (range roughly -10..+10), not 0..1. The qmd consumer (`store.ts` blend formula + `--min-score 0.3` default) assumes 0..1, so without normalization every blended score ends up negative and "No results found" fires on every query. Applying `σ(x) = 1/(1+e^-x)` in `RemoteLLM.rerank` before returning preserves ordering and fixes the range. Safe no-op for rerankers that already emit 0..1.

**B) `RemoteLLM.expandQuery` implementation (~100 lines).** The `expand_api_*` config slots are present but `expandQuery()` throws, forcing all expansion through `HybridLLM`'s local fallback. We implemented it as a POST to `<expand_api_url>/chat/completions` with a system prompt that gets a chat model (we run `gemma3:4b` on ollama cloud) to emit the same `lex:`/`vec:`/`hyde:` line-prefixed format that `LocalLLM.expandQuery` produces via grammar-constrained sampling. Parsing + fallback shape mirror `LocalLLM` exactly. End-to-end: 9 expansions in 1.3s against ollama-cloud, vs. ~minutes on a CPU-only host running the 1.7B local model. Happy to file as a follow-up PR if useful.

**C) Pre-flight probe at `vectorIndex()` startup (~30 lines).** When `usesRemoteEmbedding === true`, configuration mistakes currently surface only mid-batch. A 1-token probe at startup catches them before any real work. We hit this the hard way — silent fallback embedded for weeks on the wrong model. Optional, but cheap.

Operational signal: warm-query throughput averages 719ms across diverse queries (range 540–1012ms). No regressions in downstream consumers.

Kaspre:
Hi @georgelichen — I've opened tobi/qmd#705, a consolidated PR that builds directly on your `RemoteLLM`/`HybridLLM` work here, rebased onto current `main`. It layers on a few production-correctness refinements: sigmoid normalization for log-odds rerankers, a `RemoteLLM.expandQuery` implementation, a startup pre-flight embed probe, a `HybridLLM.rerank` local fallback (symmetric with your `expandQuery` fallback), and oversized-rerank split/truncate recovery. The core architecture is credited to you in the PR. Would really value your review.

georgelichen:
Thanks again, @Kaspre — and thanks for building on this work so thoughtfully, and for the attribution in #705.

I've now gone through #705 and left a review there. I really appreciate the consolidation and the production-hardening work on top of the original RemoteLLM/HybridLLM path.

georgelichen:
@tobi Could you please triage #629 against #705?

#629 is rebased onto current main at f49f927 and is still open/non-draft, but it has no formal review yet. The head is the georgelichen/qmd fork, and the CI/Nix workflows are currently action_required; only the Socket checks have completed.

Would you prefer:
1. reviewing #629 as the canonical PR, or
2. treating #705 as the successor and retiring #629?

If #629 is preferred, could you approve the workflows and assign/request one reviewer? I can handle focused findings promptly.

PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #644, opened 2026-05-14
fix: return absolute line numbers from qmd_query

Ginja:
## Problem

`qmd_query` results include a snippet header in diff format (`@@ -N,M @@`). Today N is the line within the matching chunk text, not the line in the source markdown.

An agent that wants to read more context around a hit can't use the snippet header to figure out where to read from in the source file. To recover the source line number, it has to grep the underlying file for the snippet text, then call `qmd_get`. Same behaviour on the HTTP `/query` endpoint and `qmd query` CLI.

## Root cause

`extractSnippet` takes `(body, query, maxLen, chunkPos)`. Pass the full body plus the chunk's character offset and it returns line numbers absolute in `body`. Pass just the chunk text (which is what all three call sites did) and `chunkPos` has nothing to anchor against, so the offset stays at zero and lines come out chunk-local.

PR #149 scoped `extractSnippet`'s keyword search to the best matching chunk so the keyword scan stays within the chunk the reranker selected. This fix preserves that scoping: `extractSnippet(body, query, 300, chunkPos)` slices `body[chunkPos-100, chunkPos+CHUNK_SIZE+100]` for the keyword scan, i.e. the same chunk window. Only the line offset accounting changes.

## Changes

`src/mcp/server.ts`:
- `SearchResultItem` gains `line: number` (additive)
- MCP `query` handler and HTTP `/query` endpoint now pass `r.body` +
  `r.bestChunkPos` to `extractSnippet` and emit `line` in the response
- HTTP endpoint now forwards `params.intent` (previously not passed)
- Tool description gains a one-liner about the `line` → `qmd_get` handoff

`src/cli/qmd.ts`:
- One-line change in the `querySearch` → `OutputRow` mapping:
  `body: r.bestChunk` → `body: r.body`
- Downstream `extractSnippet` calls already passed `row.chunkPos`, so
  the absolute-line math works once `row.body` is the full document

## The `--full` discrepancy

Tracing the bug surfaced something separate. `qmd query --full` was emitting the best chunk (~3.6KB), not the full document. CLI help text, README, and CLAUDE.md all say `--full` means "Output full document content."

The chunk-only behaviour came from `body: r.bestChunk` in the same `OutputRow` mapping. The sibling `vectorSearch` and `searchFTS` mappings passed `body: r.body` (full doc).

The CLI change above (`body: r.bestChunk` → `body: r.body`) makes `--full` emit the full document, which aligns with the documented behaviour as a side effect. Output is now proportional to document size: a representative PDF query went from 37KB to 7.4MB of JSON. Walltime impact is zero (see Performance below).

This is a behaviour change for anyone scripting against `qmd query --full`. CHANGELOG notes it.

## Tests

- New test in `test/mcp.test.ts`: fixture document with a marker at absolute line 301 (past the first chunk boundary at ~3600 chars). Asserts `result.line === 301` and the snippet header references the 300s, not a chunk-local single/double-digit line.
- Existing `extractSnippet` tests in `test/store.test.ts` and `test/intent.test.ts` continue to pass.

## Performance

Verified end-to-end on a 2K-document index with mixed body sizes including a `pdfs` collection (max body ~2.4MB):

| Surface | Δ median | Verdict |
|---|---|---|
| `extractSnippet` micro (5KB to 2MB body) | +0 to +0.4ms | sub-1ms absolute |
| MCP query snippet loop, markdown collections | +0.5 to +0.65ms | sub-1ms absolute |
| MCP query snippet loop, PDF collection | +2.23ms | ~0.2% of full pipeline |
| CLI `qmd query --json` walltime | within noise | dominated by ~1.27s expansion |

Tried replacing `String.prototype.split('\n')` with a hand-rolled `charCodeAt` newline counter to cut the prefix-split cost. It was 2-4x slower on every body size (V8's native split is SIMD-accelerated C++, the JS loop is interpreted). If this ever becomes hot, an option would be precomputing a line-start index at chunk-write time.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,17 @@
 
 ### Fixes
 
+- Snippet line numbers: `qmd_query` (MCP), HTTP `/query`, and `qmd query`
+  (CLI JSON output and snippet headers) now return absolute source-file
+  line numbers instead of chunk-local ones, so the `line` field can be
+  passed back to `qmd_get` as `fromLine` without a separate lookup.
+  Snippet selection remains scoped to the best matching chunk
+  (preserves #149).
+- CLI: `qmd query --full` now emits the full document body in all output
+  formats (json, csv, md, xml), restoring the documented behavior of the
+  flag. Previously it returned only the best matching chunk (~3.6KB max
+  per result). Output payload for `--full` queries is now proportional
+  to total document size.
 - Embedding: `qmd embed -c <collection>` now scopes pending-doc selection
   to the requested collection instead of embedding global pending work.
   Scoped `--force` clears only collection-owned vectors, preserves shared
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -844,6 +844,7 @@ function getDocument(filename: string, fromLine?: number, maxLines?: number, lin
       inputPath = inputPath.slice(0, -colonMatch[0].length);
     }
   }
+  if (fromLine !== undefined) fromLine = Math.max(1, fromLine);
 
   const parsedIndexPath = isVirtualPath(inputPath) ? parseVirtualPath(inputPath) : null;
   if (parsedIndexPath?.indexName) {
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -1886,6 +1886,7 @@ type OutputRow = {
   score: number;
   context?: string | null;
   chunkPos?: number;
+  chunkLen?: number;
   hash?: string;
   docid?: string;
   explain?: HybridQueryExplain;
@@ -1968,9 +1969,9 @@ function outputResults(results: OutputRow[], query: string, opts: OutputOptions)
     // JSON output for LLM consumption
     const output = filtered.map(row => {
       const docid = row.docid || (row.hash ? row.hash.slice(0, 6) : undefined);
+      const snippetInfo = extractSnippet(row.body, query, 300, row.chunkPos, row.chunkLen, opts.intent);
       let body = opts.full ? row.body : undefined;
-      const snippetInfo = !opts.full ? extractSnippet(row.body, query, 300, row.chunkPos, undefined, opts.intent) : undefined;
-      let snippet = snippetInfo?.snippet;
+      let snippet = !opts.full ? snippetInfo.snippet : undefined;
       if (opts.lineNumbers) {
         if (body) body = addLineNumbers(body);
         if (snippet) snippet = addLineNumbers(snippet);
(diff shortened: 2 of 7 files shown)


Ginja:
A few extra fixes got folded in after the PR body was written.

**Snippet line numbers commit additions:**

- `extractSnippet` guard changed from `chunkPos > 0` to `chunkPos !== undefined && chunkPos >= 0`. First-chunk hits (`bestChunkPos === 0`) were skipping chunk-scoped search and rescanning the full document.
- `r.bestChunk.length` threaded as `chunkLen` at all 7 call sites (2 in `src/mcp/server.ts`, 5 in `src/cli/qmd.ts` via `OutputRow.chunkLen`). Without it the chunk-window default was `CHUNK_SIZE_CHARS = 3600`, so short heading-bounded chunks had a window that overran into adjacent chunks.
- Added a layered `bestScore <= 0` fallback that handles two distinct cases:
  - `chunkPos === 0`: recurse with full body. The upstream chunk selector defaults `bestIdx = 0` when lexical chunk scoring finds no winner (e.g. tokens filtered to empty by the `length > 2` guard, or semantic-only matches with no lex overlap), so `chunkPos = 0` is potentially unreliable and the real match may live elsewhere.
  - `chunkPos > 0`: anchor on the chunk start. The reranker actively picked this chunk; tokens failing to match literally is most likely a tokenizer limitation (quoted phrases, FTS5 syntax like `-exclusions`, HYDE passages, semantic hits), so trust the chunk hint rather than disregarding it.
- CLI default `cli` format now honors `--full`. It was unconditionally emitting a snippet regardless of the flag, so `qmd query "..." --full` silently ignored the flag. Other formats (json/md/xml/csv) were already correct.
- CSV `--full --lineNumbers` now numbers from 1 (matches json/md/xml). Was numbering from the snippet match line, so a full-document dump for a hit at line 2791 labelled line 1 of the document as `2791:`.
- CLI JSON always emits `line`, including under `--full`.

**Negative `fromLine` in `qmd_get` (second commit):**

The original tool description recommended `fromLine = line - 20`, which yields negative values for hits in lines 1-20. Negative slice starts count from the end of the array, so top-of-file hits returned tail content or empty. Clamped in three layers (MCP `get` handler, CLI `getDocument`, defensive in `store.getDocumentBody`) and tightened the docstring to `fromLine = max(1, line - 20)`. Regression tests in `test/cli.test.ts` and `test/store.test.ts`.

tobi (approved):
Reviewed absolute line-number handling across store, CLI, MCP, and HTTP query output. CI-mode targeted tests pass on dev; no blocking issues found.

PR #648, opened 2026-05-15
perf(mcp): emit terse collection summary in serverInstructions

### LINKED ISSUE #647: MCP serverInstructions injects full collection catalogue on every session start — make it opt-in
The MCP `serverInstructions` emitted at `initialize` currently lists every collection with name, document count, and description (`dist/mcp/server.js` `buildInstructions`, lines 74-83). With ~12 collections that's ~1.5 KB injected into every session's system prompt, even if the agent never queries qmd that session.

For agents using long-lived sessions or many sessions per day across a fleet, this is repeated token cost for static information already discoverable via a tool call.

**Proposal** — either:
1. Collapse the collection list to a one-liner by default: e.g. `Collections (scope with \`collection\` parameter): name1, name2, …` plus a hint that the `status` tool returns descriptions and per-collection doc counts on demand. Optionally gate the full catalogue behind `QMD_MCP_VERBOSE_INSTRUCTIONS=1`.
2. Add a dedicated `list_collections` tool (or MCP resource) returning the catalogue on demand, and drop the per-collection lines from `buildInstructions` entirely.

Either approach keeps discoverability while letting agents that don't need qmd this session avoid the prompt cost. Happy to PR if you're open to (1) or (2) — sketch of (1) coming as a follow-up PR.

Tested against `@tobilu/qmd@2.1.0`.

rubycerebra:
Closes #647.

`buildInstructions` previously emitted one line per collection (name + doc count + description) into the MCP server prompt at `initialize`. On my setup that's ~1.5 KB for 12 collections, injected into every session whether the agent queries qmd or not.

This change emits a single comma-joined names line plus a one-line hint that the existing `status` tool returns descriptions and doc counts on demand. The `listContexts()` lookup is dropped because the per-collection descriptions are no longer rendered there.

**Before** (12 collections, ~1.5 KB):
```
Collections (scope with `collection` parameter):
  - "plans" (423 docs) — Claude Code implementation plans...
  - "films" (4347 docs) — Letterboxd watchlist...
  - ...etc...
```

**After** (~200 B regardless of count):
```
Collections (scope with `collection` parameter): plans, films, health, diarium, ...
Call the `status` tool for collection descriptions, paths, and per-collection doc counts.
```

Discoverability is preserved — `status` is already registered and returns the catalogue. Agents that actually need it can call once; agents that don't pay nothing.

## Test plan
- [x] `npm run build` clean
- [x] `npx vitest run test/mcp.test.ts` — 56/56 tests pass
- [x] Local install verified — session-start MCP block now ~200 B vs ~1.5 KB previously

Open to a different wording or a `QMD_MCP_VERBOSE_INSTRUCTIONS=1` opt-back-in if you'd rather keep the full catalogue available by env flag — happy to amend.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -106,7 +106,6 @@ function getPackageVersion(): string {
  */
 async function buildInstructions(store: QMDStore): Promise<string> {
   const status = await store.getStatus();
-  const contexts = await store.listContexts();
   const globalCtx = await store.getGlobalContext();
   const lines: string[] = [];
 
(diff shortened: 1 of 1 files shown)


tobi (approved):
Reviewed the terse MCP instructions change. Discoverability is preserved via collection names plus the status-tool hint; initialize still returns result.instructions (not serverInstructions); no verbose env needed unless users prove they depend on eager descriptions. Local checks: npm run build; CI=1 npx vitest run --reporter=verbose test/mcp.test.ts; targeted non-CI HTTP initialize smoke.

PR #666, opened 2026-05-20
Increase multi_get default maxBytes

fa1k3:
## Summary

Raises the default `multi_get` size limit from 10KB to 64KB.

## Why

The previous 10KB default skips fairly normal documentation/wiki pages unless callers know to override `maxBytes`. A 64KB default is still conservative, but covers more typical long-form docs without changing the explicit `maxBytes` override behavior.

## Changes

- Change `DEFAULT_MULTI_GET_MAX_BYTES` to `64 * 1024`.
- Use that shared constant in the MCP `multi_get` input schema instead of repeating a literal default.
- Update CLI help and docs to show the 64KB default.
- Add a regression test proving default `multi_get` no longer skips a normal wiki-sized file, while explicit low `maxBytes` still skips large files.

## Tests

- `npm run build`
- `npm run test:types`
- `npm run test:node -- test/mcp.test.ts` — 58 passed
- `npm run test:node` — 892 passed
- `npm run test:bun` — 892 passed


fa1k3:
Superseded by #702, which carries the same maxBytes change rebased onto current main with the conflict-resolution fixes.

PR #667, opened 2026-05-20
Report ignored documents distinctly

fa1k3:
## Summary

- return a distinct `excluded_by_ignore` lookup result when a requested document matches a configured collection ignore rule
- surface that reason through the MCP `get` tool, `qmd://` resource, and CLI `qmd get` instead of presenting it as a generic missing file
- route CLI `qmd get` through the shared document lookup path so missing/ignored paths no longer fall back to an unrelated same-basename file
- add regression tests for absolute-path, virtual-path, CLI ignored-path, and CLI missing-path lookups

## Why

If a file is intentionally excluded by an ignore rule, `get` currently looks the same as a typo or stale path. Returning the matched collection/path/rule makes the failure actionable without scanning the index manually.

The CLI also had a separate fallback that matched only the last path component. A request like `private/log.md` could return a different indexed `*log.md` file. Reusing the shared lookup avoids that surprising cross-path fallback.

## Tests

- `bun test test/cli.test.ts --timeout 60000 --preload ./src/test-preload.ts`
- `npm run test:node -- test/cli.test.ts`
- `bun test test/store.test.ts test/mcp.test.ts --timeout 60000 --preload ./src/test-preload.ts`
- `npm run test:node -- test/store.test.ts test/mcp.test.ts`
- `bun run build`
- `npm run test:types`

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -21,8 +21,7 @@ import {
   removeCollection,
   renameCollection,
   findSimilarFiles,
-  findDocumentByDocid,
-  isDocid,
+  findDocument,
   matchFilesByGlob,
   getHashesNeedingEmbedding,
   clearAllEmbeddings,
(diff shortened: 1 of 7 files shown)


fa1k3:
Superseded by #701 — rebased onto current `main` (this PR's head branch had been deleted, detaching the head so it couldn't be updated in place). Same change, conflict-resolved + full suite green. Closing in favor of the fresh branch.

PR #677, opened 2026-05-24
feat(mcp): add --host flag (and QMD_HOST env) to the HTTP server

djalmajr:
## Problem

`startMcpHttpServer` hardcodes `httpServer.listen(port, "localhost", …)`, so the HTTP MCP server only accepts loopback connections. When `qmd mcp --http` runs inside a container, a Kubernetes liveness/readiness probe connects from the pod IP (e.g. `10.42.x.x`), not `127.0.0.1` — the bind refuses it and the probe fails, so the pod never goes ready.

Today the only workaround is to patch the published `dist/mcp/server.js` at image-build time:

```dockerfile
sed -i 's@listen(port, "localhost"@listen(port, process.env.QMD_HOST || "localhost"@g' "$QMD_DIST"
```

which is brittle and breaks on every release.

## Change

Add a `--host` flag to `qmd mcp --http`, with a `QMD_HOST` environment-variable fallback:

```
host = options.host ?? process.env.QMD_HOST ?? "localhost"
```

- **Default is unchanged** (`localhost`) — no behavior change out of the box.
- `--host 0.0.0.0` (or `QMD_HOST=0.0.0.0`) binds all interfaces for container/probe reachability.
- The flag is forwarded to the detached process in `--daemon` mode and reflected in the startup log / `Started on …` message.

Files:
- `src/mcp/server.ts` — `startMcpHttpServer` resolves and binds `host`; JSDoc + listening log updated.
- `src/cli/qmd.ts` — `--host` flag parsed, forwarded to daemon spawn, shown in logs.
- `README.md` — documents `--host` / `QMD_HOST`.

## Verification

- `npm run test:types` — passes.
- `test/mcp.test.ts` — 57/57 pass (existing HTTP test uses the default and still binds `localhost`).
- Functional smoke:
  ```
  $ qmd mcp --http --host 0.0.0.0 --port 18234
  QMD MCP server listening on http://0.0.0.0:18234/mcp
  $ lsof -nP -iTCP:18234 -sTCP:LISTEN
  node … TCP *:18234 (LISTEN)        # * = all interfaces
  $ curl -s -o /dev/null -w '%{http_code}' http://127.0.0.1:18234/health
  200
  ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -120,12 +120,17 @@ By default, QMD's MCP server uses stdio (launched as a subprocess by each client
 # Foreground (Ctrl-C to stop)
 qmd mcp --http                    # localhost:8181
 qmd mcp --http --port 8080        # custom port
+qmd mcp --http --host 0.0.0.0     # bind all interfaces (e.g. container probes)
 
 # Background daemon
 qmd mcp --http --daemon           # start, writes PID to ~/.cache/qmd/mcp.pid
 qmd mcp stop                      # stop via PID file
 qmd status                        # shows "MCP: running (PID ...)" when active
 ```
+
+The server binds to `localhost` by default. Pass `--host` (or set the `QMD_HOST`
+environment variable) to override — `--host 0.0.0.0` is useful when the server
+runs in a container and a liveness probe connects from a non-loopback address.
 
 The HTTP server exposes two endpoints:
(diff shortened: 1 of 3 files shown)


designermonkey:
Hello @tobi , do you have a plan for when this will be in a release?

PR #687, opened 2026-05-27
fix: prevent macOS Metal JSON query aborts after successful output

sk-holmes:
## Problem

On macOS Metal, `qmd query --json` and `qmd deep-search --json` can produce valid JSON and then abort during native llama.cpp cleanup. That leaves callers with a non-zero exit even though the command already completed successfully.

This mainly breaks machine-readable JSON workflows. The output is valid, but the process lifecycle reports failure.

## Fix

This adds a macOS-only supervised exit path for GPU-backed JSON `query` and `deep-search` runs.

The parent process runs the query in a child process and accepts success only when both conditions are true:

- stdout parses as valid JSON
- the child writes a success sentinel on a side-channel fd after output is flushed

The child then terminates before native Metal finalizers can abort after successful output. Real query failures still return through the normal failure path.

I also tightened the GPU env handling so the supervisor only runs when GPU use is actually enabled, and the CLI reuses the same GPU-disabled semantics as the LLM layer.

## Verification

- `node scripts/build.mjs`
- `node ./node_modules/typescript/bin/tsc -p tsconfig.build.json --noEmit`
- full Node/Vitest suite: 24 files passed, 824 passed, 73 skipped
- real macOS fixture checks with `QMD_FORCE_CPU=0 QMD_LLAMA_GPU=metal`:
  - `qmd query "metal cleanup" --json -c docs`
  - `qmd deep-search "metal cleanup" --json -c docs`

## Notes

The workaround is intentionally narrow: Darwin only, JSON only, `query`/`deep-search` only, non-child process only, and disabled for explicit CPU/no-GPU runs.

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -4,8 +4,9 @@ import fastGlob from "fast-glob";
 import { execSync, spawn as nodeSpawn } from "child_process";
 import { fileURLToPath } from "url";
 import { basename, dirname, join as pathJoin, relative as relativePath, resolve as pathResolve } from "path";
+import { constants as osConstants } from "os";
 import { parseArgs } from "util";
-import { readFileSync, readdirSync, realpathSync, statSync, existsSync, unlinkSync, writeFileSync, openSync, closeSync, mkdirSync, lstatSync, rmSync, symlinkSync, readlinkSync, copyFileSync } from "fs";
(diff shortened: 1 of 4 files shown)


sk-holmes:
Looks like it was fixed in 2.5.3

PR #693, opened 2026-05-31
fix(mcp): guard reranker lazy-init against concurrent callers

### LINKED ISSUE #682: Reranker `Object is disposed` error on first `query` call to fresh MCP server
## Description

The first `mcp__plugin_qmd_qmd__query` call into a freshly-started `plugin:qmd:qmd` MCP server, when invoked with `rerank: true` (the implicit default), throws:

```
Object is disposed
```

The error fires at the reranker stage. The MCP server's `status` endpoint stays healthy throughout. The same call with `rerank: false` returns rank-1 hits instantly. Verified twice in a row on 2026-05-25 across two cold MCP server starts.

This looks like a GGUF context disposal race on the reranker model — the reranker appears to be initialized lazily and the first request hits it mid-construction. Likely sibling to a separate embed-vs-MCP contention symptom we observed in the same session, where a full `qmd embed -f` against a new collection orphans chunks with `LLM session expired before embedding chunk` errors while the MCP server is resident. Both look like node-llama-cpp model-context contention; a holistic look at model-context lifecycle in the MCP server might fix both.

## Reproduction

1. Close all running Claude Code sessions so the MCP server is not warm.
2. Open Claude Code in a session that has the qmd plugin enabled.
3. In the first turn of the session, invoke the MCP tool with the default (rerank-enabled) shape — e.g.:

```json
{
  "tool": "mcp__plugin_qmd_qmd__query",
  "searches": [{"type": "lex", "query": "test"}],
  "rerank": true,
  "intent": "test query"
}
```

## Expected

A normal hybrid + rerank result, or — if the reranker is intentionally lazy-loaded — a brief delay while the model warms followed by a successful rank-1 hit.

## Actual

```
Error: Object is disposed
```

The MCP `status` endpoint stays healthy. Same call with `rerank: false` returns the expected rank-1 hit in sub-second time.

## Workaround

Always pass `rerank: false` explicitly on the FIRST MCP `query` of a session. After at least one successful `rerank: false` query has run against the warm context, `rerank: true` becomes safe. If `rerank: true` fails despite the warm-context precondition, fall back to `rerank: false` for the rest of the session — do not retry twice in a row.

## Suggested fix

Eager-load the reranker GGUF at MCP server startup, OR guard the lazy-init path against double-entry (the disposal error suggests the first caller's context is being torn down while a second caller is mid-flight). Worth checking the same code path in the embed pipeline — same family of error there.

## Environment

- OS: Windows 11 Pro (10.0.26200)
- qmd: v2.5.2 (build 376edbe)
- Reranker: `qwen3-reranker-0.6b-q8_0.gguf` (639 MB)
- Device: `QMD_FORCE_CPU=1` (vulkan-on-iGPU crashes — separate issue)
- MCP server: `plugin:qmd:qmd` auto-wired by plugin install

tobi:
Fixes #682

## Root cause

`ensureRerankContexts()` had no concurrent-init guard. The embed path already had `embedContextsCreatePromise` to serialize parallel callers, but the reranker never got one.

On a cold MCP server start, two overlapping `query` calls with `rerank: true` both see `rerankContexts.length === 0`, both enter the init block, and the inactivity timer disposes the half-built context of whichever caller loses the race → `Error: Object is disposed`.

## Fix

Add `rerankContextsCreatePromise` field in `LlamaCpp` — identical mutex pattern to `embedContextsCreatePromise`. Concurrent cold-start callers now await the in-flight promise instead of racing.

Also clear it in `dispose()`.

## Tests added

- `test/mcp.test.ts` — new cold-start `rerank: true` test: sends `rerank: true` as the first tool call to a fresh HTTP server instance (in `skipIf(CI)` block, needs real LLM models)
- `test/llm.test.ts` — unit test for the concurrent-init race: instruments `createRankingContext` to count calls, fires two parallel `rerank()` calls at cold start, asserts contexts created exactly once
- `test/smoke-install.sh` — `run_mcp_smoke()`: spawns `qmd mcp` as a subprocess, drives JSON-RPC over stdio (initialize → notifications/initialized → query with `rerank: false`), asserts valid responses; wired into `run_runtime_matrix()` for all three install scenarios

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixes
+
+- MCP: fix `Object is disposed` error on the first `query` call with `rerank: true` against a freshly-started MCP server. The reranker lazy-init path lacked the concurrent-init promise guard that the embed path already had; two overlapping cold-start callers would both enter the init block, and the one to finish first would dispose the other's half-built context. Fixed by adding `rerankContextsCreatePromise` (mirrors `embedContextsCreatePromise`) so concurrent callers await the in-flight promise instead of racing. #682
+- Tests: add cold-start `rerank: true` MCP query test, and add MCP stdio smoke (initialize + query via JSON-RPC subprocess) to `smoke-install.sh`.
+
 ## [2.5.2] - 2026-05-22
 
 ### Fixes
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -1037,6 +1037,12 @@ export class LlamaCpp implements LLM {
    */
   private embedContextsCreatePromise: Promise<LlamaEmbeddingContext[]> | null = null;
 
+  // Promise guard for rerank context creation — mirrors embedContextsCreatePromise.
+  // Without this, two concurrent first-callers both see rerankContexts.length === 0,
+  // both start building contexts, and the disposal of one caller's half-built
+  // context causes "Object is disposed" in the other caller.
+  private rerankContextsCreatePromise: Promise<Awaited<ReturnType<LlamaModel["createRankingContext"]>>[]> | null = null;
+
   private async ensureEmbedContexts(): Promise<LlamaEmbeddingContext[]> {
     if (this.embedContexts.length > 0) {
       this.touchActivity();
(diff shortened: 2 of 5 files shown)


tobi:
Superseded by #874 on current main (this draft is stale against post-#782 `llm.ts`). Closing.

PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #694, opened 2026-05-31
fix(build): remove shell:true from spawnSync so paths with spaces work on Windows

### LINKED ISSUE #681: scripts/build.mjs mangles Windows process.execPath containing a space — silent install failure
## Description

On Windows installs where Node.js is at the default `C:\Program Files\nodejs\node.exe`, the `claude plugin install qmd@qmd` flow exits 0 but never produces a `dist/` directory. `bin/qmd` then reports "not built: missing dist/" and the plugin is unusable.

Root cause: `scripts/build.mjs` invokes `spawnSync(process.execPath, [...], { shell: true })`. With `shell: true`, the path is handed to `cmd.exe` unquoted; `cmd.exe` splits on the space at `Program Files` and tries to execute `C:\Program` instead. The spawn fails but the exit code does not propagate, so `npm run build` reports success and the install pipeline moves on.

## Reproduction

1. Windows machine with Node.js installed at `C:\Program Files\nodejs\node.exe` (default installer path).
2. `claude plugin install qmd@qmd`
3. Open the plugin cache (e.g. `~/.claude/plugins/cache/qmd/qmd/0.1.0/`).
4. Confirm `dist/` is absent and `qmd --version` fails with "not built".

## Expected

`npm run build` produces `dist/` OR exits non-zero with a clear "cannot execute node at <path>" message.

## Actual

`npm run build` exits 0 silently. `dist/` is never produced. Plugin install pipeline reports success.

## Workaround

```
cd <plugin-cache>/qmd/0.1.0
npm install
npx tsc -p tsconfig.build.json
```

This bypasses `build.mjs` entirely. After running this, `bin/qmd --version` works.

## Suggested fix

Either:
- Pass `process.execPath` through `spawnSync` without `shell: true` (the standard Node pattern), OR
- Quote `process.execPath` (e.g. `JSON.stringify(process.execPath)`) before handing it to a shell-invoked spawn, OR
- Replace the custom build invocation with a direct `npx tsc -p tsconfig.build.json` call (the workaround that works today).

## Environment

- OS: Windows 11 Pro (10.0.26200)
- Node.js: at `C:\Program Files\nodejs\node.exe`
- qmd: v2.5.2 (build 376edbe)
- Install command: `claude plugin install qmd@qmd`

## Cross-references

- Same trap will fire on any plugin whose build script uses unquoted `process.execPath` under `shell: true`. Worth checking `scripts/build.mjs` patterns across other plugins shipping source-only via Claude Code.

tobi:
Fixes #681

## Root cause

`scripts/build.mjs`'s `run()` helper passed `shell: process.platform === 'win32'` to `spawnSync`. On Windows, this hands the command to `cmd.exe` unquoted, so a Node install at `C:\Program Files\nodejs\node.exe` gets split at the space — `cmd.exe` tries to execute `C:\Program` and fails silently (exit code not propagated). Result: `dist/` is never produced, but `npm run build` reports success.

## Fix

`shell: false`. `spawnSync` with a binary path + args array has never needed shell mode — that's only for passing a shell command string. The binary is invoked directly.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- **build**: Remove `shell: true` from `spawnSync` in `scripts/build.mjs`. On
+  Windows, passing `shell: true` caused `cmd.exe` to misparse `process.execPath`
+  when the Node.js installation path contains spaces (e.g.
+  `C:\Program Files\nodejs\node.exe`). Since `run()` always receives a real
+  binary path + args array, no shell is needed — `spawnSync` can spawn
+  executables directly. (#681)
+
 ## [2.5.3] - 2026-05-28
 
 ### Features
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -10,7 +10,7 @@ function run(command, args, options = {}) {
   const result = spawnSync(command, args, {
     cwd: root,
     stdio: "inherit",
-    shell: process.platform === "win32",
+    shell: false,
     ...options,
   });
   if (result.status !== 0) {


sv789:
Tested this change on Windows 11 Pro (10.0.26200), Node v24.16.0 at the default `C:\Program Files\nodejs\node.exe`, repo at v2.6.3 (`e428df7`):

- **Without it:** `node scripts/build.mjs` fails with `'C:\Program' is not recognized as an internal or external command`, exit 1, no `dist/` produced — cmd.exe splits the unquoted exec path at the space, same mangling as #681.
- **With it:** build exits 0, `dist/` is created, and `node dist/cli/qmd.js --version` prints `qmd 2.6.3 (e428df7)`.

`run()` is only ever called with `process.execPath` plus a JS file path, so nothing here needs a shell. This currently blocks building from source on any default Windows Node install — would be great to see it land.

tobi:
Superseded by #858 — same fix rebased onto current main (CHANGELOG-only conflict).

tobi:
Closed as superseded by #858.

PR #695, opened 2026-05-31
fix(embed): make session maxDuration env-configurable via QMD_EMBED_MAX_DURATION_MS

### LINKED ISSUE #673: embed: hardcoded 30-minute session maxDuration aborts large-corpus embeddings (store.js:1377)
### Summary

The `generateEmbeddings` LLM session in `dist/store.js:1377` has a hardcoded `maxDuration: 30 * 60 * 1000` (30 minutes). On large corpora that take longer than 30 min to fully embed, the session aborts mid-run via `abortController`, `removeIncompleteEmbeddings()` purges partially-embedded documents, and the user is forced into many resume runs that may never converge.

### Reproduction (qmd 2.5.1, Apple Silicon CPU)

Corpus: ~6,400 markdown docs across two collections (mostly Chinese, ~66,000 chunks total after re-chunk).
Model: `Qwen3-Embedding-0.6B-Q8_0` via `QMD_EMBED_MODEL`.

| Run | Cmd | Wall time | Result | `needsEmbedding` after |
|-----|-----|-----------|--------|------------------------|
| 1 | `qmd embed -f` | **30:02** | "Session expired — skipping 2047 remaining chunks". 8,382 chunks embedded. | 5,594 |
| 2 | `qmd embed` (resume) | **30:02** | "Session expired — skipping 1279 remaining chunks". 9,436 chunks embedded. | 5,364 |
| 3 | `qmd embed` after patching `maxDuration` to `4 * 60 * 60 * 1000` | **3h 3m** | Clean completion: "Embedded 48023 chunks from 5364 documents". No expiry. | **0** |

Why resume doesn't converge: doc-level completion requires *all* chunks for a content hash to finish in a single session before `removeIncompleteEmbeddings()` keeps them. Each 30-min run re-chunks the not-fully-done set, persists ~8-9k chunks, then aborts — so vectors grow but the doc-level "done" set only advances by the docs that happened to fit entirely within a 30-min window. At the observed rate (~230-323 content hashes flipped to done per run), convergence on this corpus would take ~17-24 additional runs.

### Memory note

The `2.1.0` fix (#500 / #393) holds. Memory stayed bounded at ~3 GB throughout the 3h+ patched run on Apple Silicon. The 30-min cap appears to be over-defensive given that the underlying memory leak is fixed.

### Workaround (used in production)

Monkey-patch `dist/store.js:1377`:

```js
// before
}, { maxDuration: 30 * 60 * 1000, name: 'generateEmbeddings' });

// after
}, { maxDuration: 4 * 60 * 60 * 1000, name: 'generateEmbeddings' });
```

Survives until next `npm install -g @tobilu/qmd`.

### Proposed fix

Make it env-configurable, in the spirit of `QMD_EMBED_CONTEXT_SIZE` / `QMD_EXPAND_CONTEXT_SIZE`:

```js
maxDuration: Number(process.env.QMD_EMBED_MAX_DURATION_MS ?? 30 * 60 * 1000)
```

Or add a `--max-duration <ms>` flag on `qmd embed`. The 30-min default is sensible as a safety net; large corpora need an escape hatch.

### Environment

- qmd 2.5.1 (installed via `npm install -g @tobilu/qmd`)
- Node — system default via Homebrew
- macOS, Apple Silicon
- Embedding model: `hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf`
- Reranker: default `qwen3-reranker-0.6b`

tobi:
Fixes #673

## Root cause

`generateEmbeddings` session had a hardcoded `maxDuration: 30 * 60 * 1000`. On large corpora (e.g. 6,400 docs / 66,000 chunks), the 30-minute ceiling fires mid-run, `removeIncompleteEmbeddings()` purges partial work, and resume runs never converge — each run only flips ~230–323 docs to "done" so convergence takes 17–24 additional runs.

The 30-minute cap was added as a guard against the old node-llama-cpp memory leak (fixed in #500/#393). With that leak gone, the defensive cap just hurts large corpora.

## Fix

Make it env-configurable:

```ts
maxDuration: Number(process.env.QMD_EMBED_MAX_DURATION_MS ?? 30 * 60 * 1000)
```

Default stays 30 minutes (no behavior change for existing users). Large-corpus users set `QMD_EMBED_MAX_DURATION_MS=14400000` (4h) or remove the ceiling entirely with a very large value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,18 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- The embed session `maxDuration` is now env-configurable via
+  `QMD_EMBED_MAX_DURATION_MS` (default: 30 min). This prevents large-corpus
+  embeddings from being aborted by the hardcoded 30-minute ceiling (#673).
+- **build**: Remove `shell: true` from `spawnSync` in `scripts/build.mjs`. On
+  Windows, passing `shell: true` caused `cmd.exe` to misparse `process.execPath`
+  when the Node.js installation path contains spaces (e.g.
+  `C:\Program Files\nodejs\node.exe`). Since `run()` always receives a real
+  binary path + args array, no shell is needed — `spawnSync` can spawn
+  executables directly. (#681)
+
 ## [2.5.3] - 2026-05-28
 
 ### Features
--- a/src/store.ts
+++ b/src/store.ts
@@ -1824,7 +1824,7 @@ export async function generateEmbeddings(
     }
 
     return { chunksEmbedded, errors: activeErrorCount(), failures: failureList() };
-  }, { maxDuration: 30 * 60 * 1000, name: 'generateEmbeddings' });
+  }, { maxDuration: Number(process.env.QMD_EMBED_MAX_DURATION_MS ?? 30 * 60 * 1000), name: 'generateEmbeddings' });
 
   return {
     docsProcessed: totalDocs,


mendelcyprys:
This looks superseded — `main` solved it with a CLI flag instead of an env var.

`962fa6e` added `qmd embed --timeout <minutes>` (`0` = no limit, default 30), wired through
the arg parser, `--help`, and `generateEmbeddings`. Alongside it the literal became
`DEFAULT_EMBED_MAX_DURATION_MS` (`src/store.ts:97`) with a per-call `options.maxDurationMs`
override, so CLI, SDK, and MCP callers are all covered. Shipped in v2.8.3.

This PR currently reports `CONFLICTING`, and both hunks conflict with that refactor: the
call site it patches no longer contains the literal, and `CHANGELOG.md` has moved on.
Rebasing would add a second mechanism for the same knob rather than fixing anything still
broken.

One reason not to revive the env-var approach as written: `Number(process.env.X ?? 30 * 60 * 1000)`
evaluates to `NaN` for a malformed value, and since `NaN > 0` is false, the timer is never
armed — so a typo would silently disable the cap entirely rather than erroring. The
`--timeout` parser rejects non-numeric and negative input explicitly.

Probably safe to close this and #673's cap portion.

PR #703, opened 2026-06-02
Add path filtering and explain metadata to search APIs

fa1k3:
## Summary

Adds two opt-in capabilities to the search APIs (CLI `query`/`search`, the MCP `query` tool, and the SDK `store.search`):

1. **`pathPrefix` folder filtering** — restrict results to a folder-prefix within the selected collection(s), e.g. `--path stack/` / `pathPrefix: "stack"`.
2. **`explain` score breakdown** — opt-in per-result retrieval trace (`scoreType`, `backendSources`, `ftsScores`, `vectorScores`, RRF trace, `rerankScore`, `blendedScore`).

Addresses the folder-scoped search request in #6.

## Details

- `pathPrefix` is normalized (slash/whitespace, leading/trailing `/`) and applied as a `COLLATE BINARY` range predicate (`>= prefix/ AND < successor`) using a folder-boundary-safe lexicographic successor; fully parameterized. Folder-only by design — a file-like path matches no documents.
- Candidate filtering is **bounded**: FTS/vector overfetch `limit * 10` candidates before the prefix is applied — the same trade-off as the existing collection filter. Highly selective prefixes can therefore under-return; this is intentional MVP behavior, documented in code, not hidden filtering. If stricter recall is wanted I'm happy to follow up with a pre-filterable candidate source.
- Under `explain` the CLI/JSON and MCP `score` is emitted at full precision so it matches `explain.blendedScore`.

## Notes

- Rebased onto current `main` (v2.6.3).
- The earlier `original_path` round-trip machinery that previously lived on this branch has been **removed** to keep the PR focused on path filtering + explain. The shared lookup paths (`findDocument`/`findDocuments`/`multiGet`) are unchanged from `main` apart from the two features above.

## Testing

- `npm run test:types`, vitest (`test:node`) and bun (`test:bun`) suites pass.
- `npm run build` passes.
- New tests: `test/path-filter.test.ts` (pathPrefix range/normalization/injection/folder-boundary) and `test/score-breakdown.test.ts` (explain).

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -1999,6 +1999,7 @@ type OutputOptions = {
   minScore: number;
   all?: boolean;
   collection?: string | string[];  // Filter by collection name(s)
+  path?: string;  // Filter by folder-prefix within collection(s)
   lineNumbers?: boolean; // Add line numbers to output
   explain?: boolean;     // Include retrieval score traces (query only)
   context?: string;      // Optional context for query expansion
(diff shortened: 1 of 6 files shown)


fa1k3:
Closing - same request as #6, which is already closed. For a single-vault setup the collection model covers scoping a search to a subtree, so I don't have a case for carrying a separate in-collection path filter across the FTS/vector/hybrid paths - especially one I'm not actually relying on.

PR #669, opened 2026-05-20
Boost title matches in hybrid search

fa1k3:
## Summary
- add a capped title-match boost for hybrid and structured search candidate ordering
- keep final scores bounded while exposing the applied title boost in `--explain`
- add regression tests for ordering, explain consistency, exact term matching, and common technical separators

## Why
Title matches are strong evidence for lookup-style queries. The boost is applied to candidate ordering before final scoring, so it can recover relevant title hits without changing the final score equation or producing scores above 1.0.

## Tests
- `bun test test/store.helpers.unit.test.ts test/store.test.ts --timeout 60000 --preload ./src/test-preload.ts`
- `npm run test:node -- test/store.helpers.unit.test.ts test/store.test.ts`
- `bun run build`
- `npm run test:types`

PR #704, opened 2026-06-02
Boost title matches in hybrid search

fa1k3:
> Supersedes #669 — rebased onto current `main` (re-opened from a network-linked fork because the original head branch/fork was detached and couldn't be updated in place). Same change; full test suite green.

## Summary
- add a capped title-match boost for hybrid and structured search candidate ordering
- keep final scores bounded while exposing the applied title boost in `--explain`
- add regression tests for ordering, explain consistency, exact term matching, and common technical separators

## Why
Title matches are strong evidence for lookup-style queries. The boost is applied to candidate ordering before final scoring, so it can recover relevant title hits without changing the final score equation or producing scores above 1.0.

## Tests
- `bun test test/store.helpers.unit.test.ts test/store.test.ts --timeout 60000 --preload ./src/test-preload.ts`
- `npm run test:node -- test/store.helpers.unit.test.ts test/store.test.ts`
- `bun run build`
- `npm run test:types`


fa1k3:
Closing this - on a closer look, `main` already weights titles heavily (BM25 title weight 4.0 vs body 1.0, and the title is included in the embedded text used for vector search), so the premise that hybrid search ignores titles doesn't really hold. I'd rather not stack a third, correlated title signal on top without a benchmark showing genuine under-ranking of title matches. Happy to revisit with a small fixture + numbers if that turns out to be wrong.

Issue #549, opened 2026-04-10
Embedding session expires after exactly 30 minutes on CPU-only macOS (llama.cpp context lifetime)


bradleylab:
**Environment**
- Platform: macOS 15 (darwin-arm64, Apple Silicon)
- Node: 22.x
- QMD: 2.1.0 (@tobilu/qmd)
- Backend: CPU-only (no Metal acceleration enabled)

**Behavior**
`qmd embed` throws "Session expired -- skipping N remaining chunks" at exactly 30m 0s of wall-clock runtime, regardless of how many chunks remain. Each pass embeds approximately 33,000 chunks before hitting the timeout. The cutoff is deterministic: it fires at exactly 30 minutes every time.

**Workaround**
Re-running `qmd embed` is idempotent — it resumes from where it stopped. A vault with ~71k chunks required three passes to complete initial indexing.

**Hypothesis**
The timeout looks like a hard 30-minute session lifetime on the llama.cpp embedding context, possibly surfacing from node-llama-cpp. It may be a `ctx_lifetime` or keepalive parameter that defaults to 1800s.

**Impact**
Nightly LaunchAgent runs (`qmd update && qmd embed`) work fine for small daily deltas, but initial indexing of large vaults requires manual multi-pass runs.

domschl:
Happens also on Linux with CUDA (qmd 2.1.0, node 22.22.2, Arch Linux, CUDA backend for `hf_ggml-org_embeddinggemma-300M-Q8_0.gguf`):

```
Session expired — skipping remaining document batches
██████████████████████████████ 100%                                    

✓ Done! Embedded 235265 chunks from 3096 documents in 30m 0s
```

17% completed.

domschl:
Duplicate with:
- https://github.com/tobi/qmd/issues/124

Proposed fix:
- https://github.com/NathanSkene/qmd/commit/690ce6128fc34c9f75dd125d2e2ac555968ea6c0

Issue #124, opened 2026-02-06
qmd embed fails after 30 minutes on large collections (session maxDuration timeout)


thiagoghisi:
## Environment
- **OS:** macOS 15.5 (Apple Silicon, Mac Mini)
- **Bun:** v1.3.x
- **QMD:** latest from npm
- **Model:** embeddinggemma

## Issue

`qmd embed` on a large collection (50K+ documents, 164K chunks) succeeds for the first ~30 minutes, then **every remaining chunk fails** with `SessionReleasedError: LLM session has been released or aborted`. The process continues to 100% but only logs errors — no new embeddings are created after the 30-minute mark.

### Observed behavior

```
Embedding 50679 documents (164197 chunks, 315.5 MB)
Model: embeddinggemma

█████░░░░░░░░░░░░░░░░░░░░░░░░░  15% 25696/164197 27.6 KB/s ETA 2h 44m
⚠ Error embedding "file.md" chunk 0: SessionReleasedError: LLM session has been released or aborted
⚠ Error embedding "file.md" chunk 1: SessionReleasedError: LLM session has been released or aborted
... (138,501 identical errors)

✓ Done! Embedded 25696 chunks from 50679 documents in 30m 4s (179.0 KB/s)
⚠ 138501 chunks failed
```

**Key observations:**
- Exactly **30m 4s** runtime — matching the 30-minute timeout
- Embedded chunk count froze at 25,696 and never increased
- All 138,501 remaining chunks failed with the same `SessionReleasedError`
- The error-only phase logged 138K errors but embedded 0 additional chunks

## Root Cause

In `src/qmd.ts` line ~1564, the embed command wraps all work in a single session with a 30-minute hard timeout:

```typescript
await withLLMSession(async (session) => {
  // ... all embedding work for ALL chunks ...
}, { maxDuration: 30 * 60 * 1000, name: 'embed-command' });
```

In `src/llm.ts` line ~1042, `maxDuration` sets a `setTimeout` that aborts the session:

```typescript
const maxDuration = options.maxDuration ?? 10 * 60 * 1000;
if (maxDuration > 0) {
  this.maxDurationTimer = setTimeout(() => {
    this.abortController.abort(new Error(`Session exceeded max duration`));
  }, maxDuration);
}
```

When the ETA exceeds 30 minutes (common for large collections), the timer fires and aborts the session. The fallback per-chunk retry loop (line ~1611) also fails because it reuses the same aborted session.

[...]

PR #708, opened 2026-06-05
feat(embed): add --timeout to override the 30-minute session cap

yaanfpv:
## What this does

`qmd embed` runs inside a session capped at a hardcoded 30 minutes. When the cap is hit, the remaining document batches are skipped and you re-run `qmd embed` to continue. On a large corpus that turns one index into several supervised restarts, which is the behavior reported in #549 (and #124).

This adds an opt-in `qmd embed --timeout <minutes>` so you can set the cap yourself:

- `qmd embed --timeout 120` gives a big index a couple of hours to finish in one pass
- `qmd embed --timeout 0` removes the cap entirely
- omit it and the default stays 30 minutes, so nothing changes for existing users

## How

The flag is parsed in the embed command and threaded through `vectorIndex` into `generateEmbeddings` as a new `EmbedOptions.maxDurationMs`, which feeds the existing `withLLMSession({ maxDuration })`. A value of 0 or less disables the cap, which `withLLMSession` already supports. The default lives in a named `DEFAULT_EMBED_MAX_DURATION_MS` (30 min). The skip-and-resume behavior when the cap is reached is unchanged.

## Testing

- New unit test: `generateEmbeddings` stops early when `maxDurationMs` is exceeded (uses the existing fake-LLM harness).
- Existing embed tests pass; `npm run build` is clean.
- Manual: `--timeout 120`, `--timeout 0`, and invalid input (`abc`, negative) all behave correctly; `--help` documents it.

Related: #549, #124, #378.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Added
+
+- `qmd embed --timeout <minutes>` overrides the embed session's max duration
+  (previously hardcoded to 30 minutes). Use a larger value to let a big index
+  finish in one run, or `--timeout 0` to remove the cap entirely. When the cap is
+  reached, remaining document batches are skipped as before, so re-running
+  `qmd embed` continues where it left off.
+
 ### Documentation
 
 - README: documented collection filtering (`-c` semantics), the `collection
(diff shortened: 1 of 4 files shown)


yaanfpv:
Rebased onto main, the changelog conflict is resolved so this is mergeable again.

Quick recap on the why: a large index can't finish embedding inside the hardcoded 30 minute session, and right now the only way past it is editing the source. This keeps the 30 minute default and the existing skip-and-resume behavior, and just adds `--timeout <minutes>` (or `0` to remove the cap) so a big first index can run in a single pass. Small surface, one CLI flag plumbed through to `generateEmbeddings`, with a unit test covering the early stop.

@tobi Would love a look when you get a chance. Happy to reshape it as a config field or env var instead of a flag if you'd prefer that.

PR #711, opened 2026-06-05
fix(db): set busy_timeout so overlapping qmd processes queue instead …

### LINKED ISSUE #710: Overlapping qmd processes crash with SQLITE_BUSY (no busy_timeout): update dies at insertContent while an embed runs
## Summary

`openDatabase()` in `src/db.ts` sets no `busy_timeout`, and SQLite's default is 0. The index is WAL with exactly one writer at a time, so whenever two qmd processes overlap, the second writer's FIRST write statement throws `SQLITE_BUSY: database is locked` instead of waiting for the lock.

## How it bites in practice

1. **`update` vs a long `embed`.** A large-corpus embed (or the post-upgrade re-embed backlog) runs for hours, committing per batch. Any `qmd update` launched meanwhile (second shell, cron job, editor integration) dies at its first content write: `insertContent` / `indexFiles` -> `SQLITE_BUSY`. The update errors out and that pass reports 0 chunks embedded.
2. **First-open schema migration vs anything.** After an upgrade, the first command to open the DB runs the one-time migration over the whole index (minutes on a multi-GB index). Any other qmd command issued during it throws `SQLITE_BUSY` at `initializeDatabase` / `DROP TRIGGER`, which looks like a corrupted install but is just lock contention.

For a tool that encourages background and scheduled refreshes of a personal corpus, two qmd processes overlapping is the normal case, not an edge case.

## Repro

```
# terminal A: large backlog
qmd embed --max-docs-per-batch 500

# terminal B, while A runs:
qmd update    # -> SQLITE_BUSY at insertContent/indexFiles
```

## Environment

- qmd 2.5.3 (3f751cd), git install
- Windows 11, index ~10 GB / ~72k docs across 4 collections
- Reproduces under BOTH runtimes (bun:sqlite and node/better-sqlite3); neither sets a busy timeout by default

## Proposed fix

Set a busy timeout on the shared open path so writers queue instead of failing:

```ts
// src/db.ts
export function openDatabase(path: string): Database {
  const db = new _Database(path) as Database;
  const raw = Number(process.env.QMD_SQLITE_BUSY_TIMEOUT_MS);
  const busyTimeoutMs = Number.isFinite(raw) && raw >= 0 ? Math.floor(raw) : 120_000;
  db.exec(`PRAGMA busy_timeout = ${busyTimeoutMs}`);
  return db;
}
```

Embeds commit per batch, so a queued writer gets the lock at the next batch boundary; 2 minutes outlasts worst-case commit pauses on a 10 GB index. The env var keeps an escape hatch (`0` = old fail-fast behavior).

Running this as a local patch: an `update` launched against a held write lock queues ~3s and succeeds where it previously threw instantly. PR ready, sending it now.

ghost:
Fixes #710.

`openDatabase()` sets no `busy_timeout` (SQLite default 0), so when two qmd processes overlap, the second writer's first INSERT/DDL throws `SQLITE_BUSY: database is locked`: an `update` dies at `insertContent`/`indexFiles` while a long `embed` runs, and the first-open schema migration after an upgrade throws at `initializeDatabase`/`DROP TRIGGER` if any routine command races it.

This sets `PRAGMA busy_timeout = 120000` on the shared open path (covers both bun:sqlite and better-sqlite3), so writers queue at batch boundaries instead of failing. `QMD_SQLITE_BUSY_TIMEOUT_MS` overrides; `0` restores the old fail-fast behavior.

Verified on Windows 11 against a 10 GB / 72k-doc index under both engines: with the pragma, an `update` launched against a held write lock queued ~3s and succeeded where it previously threw immediately. Readers are unaffected (WAL).

### DIFF
--- a/src/db.ts
+++ b/src/db.ts
@@ -65,7 +65,21 @@ if (isBun) {
  * Open a SQLite database. Works with both bun:sqlite and better-sqlite3.
  */
 export function openDatabase(path: string): Database {
-  return new _Database(path) as Database;
+  const db = new _Database(path) as Database;
+  // Queue concurrent writers instead of failing immediately: SQLite's default
+  // busy timeout is 0, so when two qmd processes overlap (a long `embed` plus
+  // an `update` from a cron job or a second shell, or the one-time schema
+  // migration racing a routine command), the second writer's first INSERT/DDL
+  // throws SQLITE_BUSY ("database is locked"). WAL mode (set in store.ts)
+  // allows exactly one writer at a time; briefly waiting for the lock is the
+  // right behavior for a CLI. Embeds commit per batch, so writers interleave
+  // at batch boundaries; 2 minutes outlasts worst-case commit pauses on a
+  // multi-GB index. Override with QMD_SQLITE_BUSY_TIMEOUT_MS (0 restores
+  // fail-fast).
+  const raw = Number(process.env.QMD_SQLITE_BUSY_TIMEOUT_MS);
+  const busyTimeoutMs = Number.isFinite(raw) && raw >= 0 ? Math.floor(raw) : 120_000;
+  db.exec(`PRAGMA busy_timeout = ${busyTimeoutMs}`);
+  return db;
 }
 
 /**


tobi:
Closing as superseded — this already landed on `main`.

`openDatabase()` now sets `PRAGMA busy_timeout` (default 120s) so overlapping `qmd` processes queue instead of throwing `SQLITE_BUSY`. Override with `QMD_SQLITE_BUSY_TIMEOUT` (milliseconds; `0` restores fail-fast). Note the env name on main is `QMD_SQLITE_BUSY_TIMEOUT`, not `QMD_SQLITE_BUSY_TIMEOUT_MS` from this PR.

Landing commits:
- 553f6078fe4a13a62edc5d9ccc24a97850c93fa2 — set busy_timeout
- 616850d15dbec9cb5b868da46a7cb2cefdd0f038 — raise default to 120s + env override
- 5496df3940a94794b3342ea16a313f9dea5ad71a — concurrent-open follow-up

Issue #710 is already closed. This branch is conflicted against current main and has no remaining unique work.

PR #433, opened 2026-03-18
feat(mcp): expose explain parameter for retrieval score traces

fxstein:
## Problem

The CLI `qmd query --explain` flag includes detailed retrieval score traces per result — FTS scores, vector scores, RRF fusion breakdown, reranker blend weights, and per-source contributions. This is a shipped v2.0.1 feature.

However, the MCP `query` tool does not expose this option. MCP clients cannot debug query quality, understand score composition, or tune search parameters without dropping to the CLI.

## Fix

Add an optional `explain` boolean parameter to the MCP `query` tool (default: `false`). When enabled, pass `explain: true` through to `structuredSearch()` and include the `HybridQueryExplain` object on each result.

Changes:
- Extend `SearchResultItem` type with optional `explain` field
- Add `explain` to MCP tool input schema
- Pass `explain` through in both MCP tool handler and REST `/query` endpoint
- Conditionally include explain data in results (only when present, keeps response compact by default)

## Testing

Built and tested locally against a 7,665 document index:

| Query | explain | Result |
|-------|---------|--------|
| `searches: [{type: "lex", query: "boot gate"}]` | `true` | ftsScores, rrf breakdown, rerankScore, blendedScore all present |
| `searches: [{type: "lex", query: "boot gate"}]` | `false` | Standard result, no explain field |
| `searches: [{type: "lex", query: "boot gate"}]` | omitted | Same as false (default) |

Example explain output:

    {
      "ftsScores": [0.88],
      "vectorScores": [],
      "rrf": { "rank": 1, "totalScore": 0.083, "contributions": [{"source": "fts", "rank": 1, "weight": 2}] },
      "rerankScore": 0.502,
      "blendedScore": 0.875
    }

Existing queries without `explain` return results unchanged — no breaking changes.

## Environment

- QMD: v2.0.1
- Platform: macOS (Apple Silicon)
- Node: v24.2.0

PR #730, opened 2026-06-13
feat(mcp): expose explain parameter for retrieval score traces

fxstein:
## What

Adds an optional `explain` boolean to the MCP `query` tool and the REST `/query` (alias `/search`) endpoint. When set, each result carries the retrieval score trace the SDK already computes — per-list FTS and vector scores, the RRF fusion breakdown, and rerank blend weights.

## Why

`SearchOptions.explain` and `HybridQueryResult.explain` already exist in the SDK (`hybridQuery()` / `structuredSearch()` build the trace via `buildRrfTrace`), but there is no way to request it from an MCP client — you have to drop down to the SDK. This plumbs the existing flag through the two MCP-facing surfaces so query quality and ranking can be inspected in place.

## Details

- New `explain` param (default `false`) on the `query` tool schema, forwarded to `store.search`.
- `explain` accepted on the REST `/query` body and forwarded likewise.
- Each result spreads `...(r.explain ? { explain: r.explain } : {})`, typed against the exported `HybridQueryExplain` — no output change when `explain` is unset.
- `tsc --noEmit` clean.

## History

Fresh, focused resubmission of #433, which was closed on 2026-05-20 during a stale-backlog purge with the note to *"open a fresh focused issue if it still reproduces on v2.5.1+"*. No `explain` parameter exists on current `main`, so the gap remains. Rebased onto current `main` (the rerank-aware `query` signature); the change is smaller now because `main` already exports `HybridQueryExplain`.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -28,6 +28,7 @@ import {
   type QMDStore,
   type ExpandedQuery,
   type IndexStatus,
+  type HybridQueryExplain,
 } from "../index.js";
 import { getConfigPath } from "../collections.js";
 import { enableProductionMode } from "../store.js";
@@ -44,6 +45,7 @@ type SearchResultItem = {
   context: string | null;
   line: number;   // Absolute line in source markdown
   snippet: string;
+  explain?: HybridQueryExplain;  // Retrieval score traces, present only when explain=true
 };
 
 type StatusResult = {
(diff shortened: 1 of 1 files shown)


fxstein:
Rebased onto current `main` (v2.6.3) to clear the merge conflict — the overlap was purely positional with #731 (plain `query` param), which landed in the same `search` tool block. The `explain` param slots in alongside `query`; the two are orthogonal. `tsc --noEmit` (test:types) passes clean. Ready for review.

fxstein:
Rebased onto current `main` (`dbfd0b4`). The branch was unmergeable and now applies cleanly.

The only substantive change from the rebase: the `query` handler is now wrapped in `track(...)`, so the `explain` parameter is threaded through that wrapper rather than the bare handler signature it was written against.

```ts
    explain: z.boolean().optional().default(false).describe(...),
  }),
},
track(async ({ query, searches, limit, minScore, candidateLimit, collections, intent, rerank, explain }) => {
```

Nothing else in the change moved. Still opt-in and still defaulting to `false`, so the response shape is unchanged unless a caller asks for the trace.

PR #734, opened 2026-06-14
fix(mcp): reap idle HTTP sessions

logicrw:
## Summary
- Add idle-session cleanup for Streamable HTTP MCP clients that do not send `DELETE /mcp`.
- Add a configurable active-session cap to prevent unbounded session growth.
- Cover both idle cleanup and oldest-session eviction with HTTP transport regression tests.

## Changes
- Tracks `lastSeen` timestamps for MCP HTTP sessions.
- Reaps idle sessions using `QMD_MCP_SESSION_TTL_MS` (default: 10 minutes, `0` disables).
- Caps active sessions using `QMD_MCP_MAX_SESSIONS` (default: 64, `0` disables).
- Clears tracking state when sessions close or the server shuts down.

## Testing
- `bun run test:types`
- `env -u CI node ./node_modules/vitest/vitest.mjs run --reporter=verbose --testTimeout 60000 test/mcp.test.ts -t "MCP HTTP Transport"`
- `bun run build`

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -32,6 +32,9 @@ import {
 import { getConfigPath } from "../collections.js";
 import { enableProductionMode } from "../store.js";
 
+const DEFAULT_MCP_HTTP_SESSION_TTL_MS = 10 * 60 * 1000;
+const DEFAULT_MCP_HTTP_MAX_SESSIONS = 64;
+
 // =============================================================================
 // Types for structured content
 // =============================================================================
@@ -70,6 +73,16 @@ type StatusResult = {
 function encodeQmdPath(path: string): string {
   // Encode each path segment separately to preserve slashes
   return path.split('/').map(segment => encodeURIComponent(segment)).join('/');
+}
+
+function readNonNegativeIntegerEnv(name: string, fallback: number): number {
+  const raw = process.env[name];
+  if (raw === undefined || raw.trim() === "") return fallback;
+
+  const parsed = Number(raw);
+  if (!Number.isSafeInteger(parsed) || parsed < 0) return fallback;
+
+  return parsed;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


junmo-kim:
Ran into the same leak and went looking for prior art, which is how I found this. One thing that might be worth guarding before it lands.

`sessionLastSeen` is stamped when a request starts and isn't refreshed while it runs, so a session's own in-flight request ages it. Since `reapSessions()` runs at the top of request handling, a session can be closed while it still has a request inside `handleRequest()`.

The idle path needs a request slower than the TTL, so it's mostly out of reach at the 10 minute default. The `max-sessions` path isn't: eviction picks the oldest `lastSeen`, and a session waiting on a slow `query` has by definition the oldest one, so the longest-running request is the first thing evicted once the cap is hit.

The failure mode isn't a clean 404. In `enableJsonResponse` mode the SDK's cleanup for a pending response drops its stream bookkeeping without resolving the promise `handleRequest()` returned, so the awaited call never settles and no response is ever written. The client is left holding an open socket with no timeout.

Here's a failing test against this branch, in the style of the two you added:

```ts
test("POST /mcp does not evict a session that still has a request in flight", async () => {
  await withMcpHttpEnv({ QMD_MCP_SESSION_TTL_MS: "0", QMD_MCP_MAX_SESSIONS: "2" }, async (targetBaseUrl) => {
    const slowId = 777;
    const busy = await mcpRequestTo(targetBaseUrl, initializeRequest(1));
    expectSessionId(busy.sessionId);

    // Make just this session's request slow, so it is still inside
    // handleRequest() while the cap evicts it.
    const originalSend = WebStandardStreamableHTTPServerTransport.prototype.send;
    WebStandardStreamableHTTPServerTransport.prototype.send = async function (
      this: WebStandardStreamableHTTPServerTransport,
      ...args: Parameters<typeof originalSend>
    ) {
      const message = args[0] as { id?: unknown } | undefined;
      if (message && typeof message === "object" && message.id === slowId) {
        await new Promise(resolve => setTimeout(resolve, 2000));
      }
      return originalSend.apply(this, args);
    };

    try {
      const inFlight = mcpRequestTo(targetBaseUrl, toolsListRequest(slowId), busy.sessionId);
      await new Promise(resolve => setTimeout(resolve, 200));

      // Two more sessions push past the cap. `busy` has the oldest lastSeen,
      // because it was stamped when its still-running request started.
      expectSessionId((await mcpRequestTo(targetBaseUrl, initializeRequest(2))).sessionId);
      expectSessionId((await mcpRequestTo(targetBaseUrl, initializeRequest(3))).sessionId);

      const timedOut = Symbol("timed out");
      const settled = await Promise.race([
        inFlight,
        new Promise(resolve => setTimeout(() => resolve(timedOut), 6000)),
      ]);
      expect(settled).not.toBe(timedOut);
    } finally {
      WebStandardStreamableHTTPServerTransport.prototype.send = originalSend;
    }
  });
}, 20000);
```

It needs `import { WebStandardStreamableHTTPServerTransport } from "@modelcontextprotocol/sdk/server/webStandardStreamableHttp.js";`. Flipping `QMD_MCP_MAX_SESSIONS` to `"0"` makes it pass, which points at the eviction rather than the delay itself.

On the fix: I first tried re-stamping `lastSeen` after `handleRequest()` returns, and that doesn't work. While the request is running `lastSeen` is still the old value, which is exactly when eviction picks the session, so the re-stamp lands too late to matter. What did work was tracking sessions that currently have a request inside `handleRequest()` and skipping them both in the idle loop and in the eviction candidate filter. A refcount rather than a flag, since overlapping requests share a transport. With that, the test above and the rest of the `MCP HTTP Transport` block passed for me.

tobi:
Closing as superseded — idle HTTP session reaping already landed on `main` via #816.

Current main expires idle MCP HTTP sessions with a background reaper (default 1h). Configure with `--session-ttl` or `QMD_MCP_SESSION_TTL` (seconds; `0` disables). Landing commit: 4c5caaa74d8d50b4ae55990e8de8d27339735420 (`feat(mcp): expire idle HTTP sessions with a configurable TTL (#816)`).

This branch is conflicted against current main. The extra `QMD_MCP_MAX_SESSIONS` cap from this PR is not on main; if that cap is still wanted, it should be a fresh PR against current main rather than this stale duplicate.

PR #752, opened 2026-06-30
Expose reranker score in MCP/REST explain output

Kyzcreig:
## Summary

- Adds optional `explain: true` support to the MCP `query` tool and HTTP REST `/query`/`/search` path.
- Exposes a top-level `rerankScore` per result when explain output is requested.
- Keeps default responses unchanged: no `explain`, no `rerankScore` unless `explain: true` is supplied.
- Maps disabled-rerank sentinel behavior to `null` instead of exposing `0` as a semantic relevance score.

## Why

QMD already computes a pure semantic reranker score inside `HybridQueryExplain.rerankScore`, but callers at the MCP/HTTP boundary only receive the blended display `score`. That blended score includes RRF/position signal, so it is useful for ranking but not for applying a semantic relevance floor.

This exposes the existing pure reranker signal without changing the default response shape.

## Test plan

- `bun install`
- `bun run test:types`
- `CI=true bunx vitest run test/mcp.test.ts --reporter=verbose --testTimeout 60000`
- `bun run build`
- `NODE_NO_WARNINGS=1 bun run test:unit`

Note: `bun run test:unit` without `NODE_NO_WARNINGS=1` failed locally only on pre-existing `test/cli.test.ts` stderr-empty assertions due to Node DEP0205 `module.register()` deprecation warnings; rerunning with warnings suppressed produced `883 pass / 80 skip / 0 fail`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Added
+
+- MCP `query` and HTTP REST `/query`/`/search` accept `explain: true` and expose
+  a top-level `rerankScore` per result: the raw LLM reranker relevance score from
+  `explain.rerankScore` (or `null` when reranking is disabled). Default responses
+  are unchanged unless `explain` is requested.
+
 ## [2.6.3] - 2026-06-24
 
 ### Added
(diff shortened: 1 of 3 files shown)


Kyzcreig:
Hi! This PR has been open for about 37 days and hasn't had a look from anyone yet. It's not a draft and currently shows as mergeable, so it should be ready for a first pass whenever someone has a moment.

No decision needed from me right now — just flagging it for visibility so it doesn't keep sitting in the queue. Happy to rebase, split it up, or add more context if that would make reviewing easier. Thanks!

PR #753, opened 2026-07-02
Fix multi-get docid resolution

### LINKED ISSUE #706: multi-get does not resolve docids despite docs/examples
## Summary

`qmd get "#docid"` resolves short docids correctly, but `qmd multi-get "#docid"` and `qmd multi-get "#docid,#other"` do not, even though the README and MCP tool description say `multi-get` / `multi_get` supports docids.

This may be either a docs mismatch or a missing docid-resolution branch in `multiGet` / `findDocuments`.

## Version / environment

- qmd: 2.5.3
- installed from npm: `npm install -g @tobilu/qmd@latest`
- macOS, Node >= 22

## Expected behavior

Given a search/get result with docid `#1c9a4c`, both of these should retrieve the document, per docs:

```sh
qmd multi-get "#1c9a4c"
qmd multi-get "#1c9a4c,#688343"
```

The README says:

```text
multi_get — Batch retrieve by glob pattern, comma-separated list, or docids
```

And the bundled QMD skill says:

```sh
qmd multi-get "#abc123,#def432" --format md
```

## Actual behavior

`get` works:

```sh
$ qmd get '#1c9a4c:1:5'
qmd://obs-projects/garde-sos-planning-system.md  #1c9a4c
---

1: ---
2: id: prj-01KT6VCQQ5QMY497FS2SBEHM9J
3: type: Project
4: created_at: 2026-06-03
5: updated_at: 2026-06-03
```

But `multi-get` does not resolve the same docid:

```sh
$ qmd multi-get '#1c9a4c' -l 2
No files matched pattern: #1c9a4c

$ qmd multi-get '#1c9a4c,#688343' --format md -l 2
File not found: #1c9a4c
File not found: #688343
```

Passing a collection path works:

```sh
$ qmd multi-get 'obs-projects/garde-sos-planning-system.md' -l 2
============================================================
File: garde-sos-planning-system.md  #1c9a4c
============================================================

1: ---
2: id: prj-01KT6VCQQ5QMY497FS2SBEHM9J
```

## Likely code path

In the published 2.5.3 build:

- `get` has an explicit docid branch around `dist/cli/qmd.js`:

```js
// Handle docid lookup (#abc123, abc123, "#abc123", "abc123", etc.)
if (isDocid(inputPath)) {
  const docidMatch = findDocumentByDocid(db, inputPath);
  ...
}
```

- `multiGet` around `dist/cli/qmd.js` only treats comma-separated entries as virtual paths or path/suffix matches. It does not appear to call `isDocid()` / `findDocumentByDocid()` for each entry.

- MCP `multi_get` calls store `findDocuments(pattern, ...)`, and `dist/store.js` similarly treats comma-separated names as qmd:///path matches, not docid matches.

## Why this matters

Agents often get docids from search results and then batch-retrieve selected hits. The docs advertise exactly this pattern, but the current behavior forces agents to call `get` repeatedly or convert docids back to paths first.

## Possible fix direction

Before path/virtual-path matching in `multiGet` / `findDocuments`, resolve each comma-separated item with the same docid logic used by `get`:

- if `isDocid(name)`, call `findDocumentByDocid(db, name)`
- use the matched document path / collection to fetch the row

Alternatively, if docid support is not intended for `multi-get`, the README and bundled skill should remove the docid examples.

dpersek:
## Summary

- Resolve docids in shared `multi_get` / SDK multiGet lookups, including comma-separated lists.
- Resolve docids in the separate CLI `multi-get` path and keep missing single-docid lookups as failures.
- Update MCP `multi_get` tool text so clients know docids are accepted.

## Why

Fixes #706.

`get` already resolves docids returned by search results, but `multi-get` treated docid inputs as unmatched patterns/paths. The docs advertise docid support for multi-get, so this makes the batch retrieval path match the documented behavior.

## Tests

- `bun test --preload ./src/test-preload.ts test/store.test.ts -t "findDocuments"`
- `bun test --preload ./src/test-preload.ts test/sdk.test.ts -t "multiGet retrieves"`
- `bun test --preload ./src/test-preload.ts test/cli.test.ts -t "retrieves (a document by docid|docids in comma-separated lists)|fails for a missing single docid"`
- `git diff --check`
- `bun run test:types`

## Scope

- In scope: docid resolution for store/SDK/MCP multi-get, CLI multi-get, MCP tool text, and focused regression tests.
- Non-goals: indexing, embedding, model loading, search ranking, database schema, MCP transport behavior, or broader retrieval refactors.

## Caveats

- The broad Bun run for `test/store.test.ts test/sdk.test.ts test/cli.test.ts` hit unrelated local LLM cache and MCP HTTP daemon failures in this sandbox, so I used focused tests for the changed behavior plus typecheck.

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -22,6 +22,7 @@ import {
   renameCollection,
   findSimilarFiles,
   findDocument,
+  findDocumentByDocid,
   matchFilesByGlob,
   getHashesNeedingEmbedding,
   clearAllEmbeddings,
@@ -40,6 +41,7 @@ import {
   parseVirtualPath,
   buildVirtualPath,
   isVirtualPath,
+  isDocid,
   resolveVirtualPath,
   toVirtualPath,
   insertContent,
@@ -1072,19 +1074,24 @@ function multiGet(pattern: string, maxLines?: number, maxBytes: number = DEFAULT
 
   // Check if it's a comma-separated list or a glob pattern
   const isCommaSeparated = pattern.includes(',') && !pattern.includes('*') && !pattern.includes('?') && !pattern.includes('{');
+  const isSingleDocid = isDocid(pattern);
 
   let files: { filepath: string; displayPath: string; bodyLength: number; collection?: string; path?: string }[];
 
-  if (isCommaSeparated) {
+  if (isCommaSeparated || isSingleDocid) {
     // Comma-separated list of files (can be virtual paths or relative paths)
-    const names = pattern.split(',').map(s => s.trim()).filter(Boolean);
+    const names = isCommaSeparated
+      ? pattern.split(',').map(s => s.trim()).filter(Boolean)
+      : [pattern.trim()].filter(Boolean);
     files = [];
     for (const name of names) {
       let doc: { virtual_path: string; body_length: number; collection: string; path: string } | null = null;
+      const docidMatch = isDocid(name) ? findDocumentByDocid(db, name) : null;
+      const lookupName = docidMatch?.filepath ?? name;
 
       // Handle virtual paths
-      if (isVirtualPath(name)) {
-        const parsed = parseVirtualPath(name);
+      if (isVirtualPath(lookupName)) {
+        const parsed = parseVirtualPath(lookupName);
         if (parsed) {
           // Try exact match on collection + path
           doc = db.prepare(`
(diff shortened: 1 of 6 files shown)


rymalia:
Tested the problem space independently before finding this PR — the fix targets the right seams. Confirming the bug this addresses still reproduces on `main` at `e428df7` in all three places the PR touches (CLI comma-list, CLI single-docid-to-glob-branch routing, SDK `findDocuments` backing MCP `multi_get`), and the scope declaration matches what we found: docid resolution is cleanly separable from the comma-list resolvers' other issues.

One thing worth double-checking in review: the CLI and SDK comma-list resolvers are independent implementations with already-divergent path semantics (collection-prefixed paths work over MCP but not CLI; both suffix-match via unanchored `LIKE '%name'`). If the docid branches are added to each separately, that's two more parallel code paths to keep in sync — worth a comment in the code, or a pointer to the follow-up issue (#759) that proposes unifying them.

PR #761, opened 2026-07-09
feat(embed): add remote OpenAI-compatible embedding backend

epruseal:
## What

`models.embed` (and `QMD_EMBED_MODEL`) can now point at one or more
OpenAI-compatible `/v1/embeddings` servers instead of a local GGUF model,
via `http[s]://host:port/v1#model-id` URIs. Multiple endpoints can be listed
for sequential fallback, each with an independent 15s unhealthy backoff,
with an optional local GGUF entry as a last-resort fallback if every
remote endpoint fails.

## Why

Not everyone wants (or has the GPU for) a local GGUF embedding model — many
already run a shared OpenAI-compatible embedding server (LM Studio, vLLM,
text-embeddings-inference, Ollama's compatible endpoint, or real OpenAI).
This lets `qmd` point at that instead of downloading and running its own
model. The design mirrors a sibling project of mine that already does
OpenAI-compatible remote embedding with multi-endpoint fallback, so this
ports that pattern into `qmd`'s existing model-resolution/config system.

## How

- New `src/remote-embed.ts`: `RemoteEmbedder` — tries each endpoint in
  order, POSTs `{model, input}` to `{apiBase}/embeddings`, L2-normalizes
  results, marks a failed endpoint unhealthy for `QMD_EMBED_HEALTH_TTL_MS`
  (default 15s) before retrying it. The model id rides in the URL fragment
  (`#model-id`) rather than in the config value itself, since fragments are
  never sent over the wire — this keeps the whole thing a single string
  that flows through the existing `embed: string` config field (now widened
  to `string | string[]` for the fallback-list case).
- `LlamaCpp.embed`/`embedBatch` route to the remote client *before* touching
  any local tokenizer/GGUF setup. A plain `hf:...`/local config takes the
  exact same code path as before (`isRemoteEmbed()` gates the new branch;
  covered by a backward-compat regression test).
- All endpoints in a fallback list must share one `#model-id` — validated
  when the config loads. That shared id is the canonical identity stored in
  `content_vectors.model` and hashed into the embedding fingerprint, so a
  chunk embedded via one endpoint is found when searching through another.
- **Local GGUF as a last-resort fallback**: if an `embed` list also contains
  a local (`hf:...`) entry, `LlamaCpp.embed`/`embedBatch` fall through to it
  when every remote endpoint fails, instead of returning null/erroring. The
  returned `model` stays the canonical remote identity even when the vector
  actually came from the local GGUF, so the fallback vector lands in the
  same fingerprint/collection as vectors written while remote was healthy.
  **Under discussion below** — a reviewer flagged that this can silently mix
  incompatible vector spaces; see the PR comments for the opt-in proposal.
- `chunkDocumentByTokens` skips the tokenize/detokenize verification pass
  for remote configs (there's no local tokenizer to call) across all three
  of its call sites — `generateEmbeddings`, legacy-fingerprint adoption, and
  `qmd doctor`'s vector-sample check — falling back to a char-based token
  estimate consistent with the existing `avgCharsPerToken` heuristic.
- Fixed a latent bug along the way: SDK/MCP vector search embedded queries
  via the process-wide default `LlamaCpp` instead of the `Store`'s
  configured one (`searchVec` had no way to reach `store.llm`). Threaded an
  `llmOverride` through the store-wrapper `searchVec` so it forwards
  `store.llm` — this is what makes config-file-based remote embedding
  actually usable outside the CLI (which happened to work already since it
  calls `setDefaultLlamaCpp`).
- `qmd doctor` reports configured remote endpoints, the shared model id, and
  whether an API key is set (never its value). `qmd pull` skips remote
  entries since there's nothing to download for them.
- CLI config write-back (`ensureModelsConfiguredForCli`) no longer collapses
  an explicit `embed` array back into a single resolved string on save.

Auth (`QMD_EMBED_API_KEY`, falls back to `OPENAI_API_KEY`), timeout
(`QMD_EMBED_TIMEOUT_MS`, default 30s), and failover TTL
(`QMD_EMBED_HEALTH_TTL_MS`, default 15s) are all env vars — never in the URI.

## Testing

- New unit tests: `RemoteEmbedder` (sequential fallback, per-endpoint TTL via
  fake timers, L2 normalization, all-endpoints-fail), `LlamaCpp` remote
  routing (fetch-mocked; asserts `node-llama-cpp` is never touched on the
  remote path), local-GGUF-fallback-on-total-remote-failure (mocked GGUF,
  identity preserved), tokenizer-free chunking, fingerprint sharing across
  two endpoints with the same model id, and a config round-trip + CLI
  regression guarding that an explicit `embed` array survives `qmd doctor`'s
  write-back unchanged.
- `tsc -p tsconfig.build.json --noEmit` is clean; the full `vitest` suite
  passes (924 tests, 0 failures).
- Manual: ran the full `qmd doctor` → `qmd embed` → `qmd vsearch` cycle
  against two real LM Studio-hosted OpenAI-compatible servers (same model,
  1024-dim), in an isolated temp index/config — including killing the first
  endpoint mid-run to confirm sequential fallback actually kicks in and the
  vector already written under the shared identity is still found via
  search through the second endpoint.

Docs updated: README ("Remote Embedding" section under Model Configuration),
`example-index.yml`, and a CHANGELOG entry.

---

The CJK FTS / `database is busy` fix that was previously bundled into this
branch has been split out into #797 and dropped from here (rebased), per
review feedback.


epruseal:
Closing for now — pulling this back to finish full verification of an additional local-GGUF-fallback-on-remote-failure path before resubmitting. Will reopen or open a fresh PR once everything (including that addition) is fully tested.

boneanxs:
@tobi Hi master, do we have plan to support this? 

+1 for this!

I really like QMD’s local Markdown indexing and MCP workflow, but having to run a local embedding model is a blocker for me. I already use a remote embedding API, so being able to point QMD at an OpenAI-compatible endpoint would make it much easier to adopt.

soundadam (changes requested):
I tested this head with Node 22 / pnpm 10: TypeScript passes and the focused suite passes (347 passed, 50 skipped).

Requesting changes because the local fallback violates embedding-space identity. When all remote endpoints fail, the PR can generate a vector with a local GGUF model and deliberately return/store it under the canonical remote model identity. The test named `falls back to local GGUF (with the canonical remote identity)` currently codifies that behavior.

The model string and vector dimension are not sufficient evidence that two backends/model artifacts produce the same vector space. A remote Transformers/ONNX model and a local GGUF quantization can differ in preprocessing, pooling, normalization, tokenizer revision, and weights. Storing both under one `content_vectors.model` / fingerprint can silently mix incompatible vectors in one index.

Please fail closed on total remote failure, or persist the actual producer backend plus immutable model artifact/preprocessing fingerprint and require a full re-embed before switching. A fallback must never masquerade as the failed remote producer.

This PR also contains unrelated CJK FTS/database changes; please split those so the embedding-provider change can be reviewed and rolled back independently.

epruseal:
Thanks for the careful review, both points are fair.

## Local GGUF fallback identity

You're right that model name + dimension is not proof of vector-space
equivalence — a remote server and a local GGUF quantization can diverge in
preprocessing/pooling/normalization/tokenizer revision even when they claim
the same id. The current implementation trusts the operator's config to
guarantee equivalence, which isn't something the code can verify.

My own use case is a homelab setup: the remote endpoint is the *same* model
weights served on a GPU box, with a GGUF quantization of those same weights
kept warm locally as a standby on a small CPU device (a Raspberry Pi) for
when the remote box is down. In that narrow case the two producers are
close enough for retrieval purposes that I was comfortable accepting the
tradeoff for myself — but I agree that's a property of *my* setup, not
something the code should assume generally. Someone pointing `embed` at
e.g. real OpenAI's API with an unrelated local GGUF as "fallback" would
silently corrupt their index exactly as you describe.

Rather than drop the fallback path entirely, would it be acceptable to make
it strictly **opt-in** — fail closed by default (today's behavior for a
remote-only config, unchanged), and only attempt the local-GGUF fallback
when the operator sets an explicit flag (e.g. `QMD_EMBED_ALLOW_LOCAL_FALLBACK=1`
or a `models.embed` entry marked explicitly, TBD on shape) acknowledging they
accept the vector-space-equivalence risk for their own config? That keeps
the default safe for everyone, while still letting an operator who
genuinely controls both ends (same weights, as in my case) opt into the
tradeoff knowingly. Open to the persist-actual-producer-and-require-re-embed
design too if you'd rather go that route instead — happy to implement
whichever you prefer.

## CJK FTS / db-busy split

Agreed, will split it into its own PR so it can be reviewed and
reverted independently of the embedding-provider change.

epruseal:
Split the CJK FTS / db-busy fix out into #797 — can be reviewed and merged independently of this one.

epruseal:
Dropped the CJK FTS / db-busy commit from this branch (rebased + force-pushed) — it's now only in #797. This PR is back to just the remote-embed feature + the local-GGUF-fallback addition.

soundadam (commented):
Re-reviewed force-pushed head `770f45b65fb438e1c082d132bceb71c93e9e14d4`.

The secondary review request is resolved: the unrelated CJK FTS / database-busy change has been removed from this branch and moved to #797. TypeScript passes, and the focused suite still passes (347 passed, 50 skipped).

The primary blocker remains unchanged. The branch still explicitly tests and implements local-GGUF vectors being returned under the canonical remote model identity:

```ts
expect(result?.model).toBe("test-embed-model");
```

An operator acknowledgement flag would not make the stored index correct. It would only acknowledge that the operator is accepting silent vector-space mixing. QMD cannot subsequently determine which vectors were produced by which tokenizer, preprocessing, pooling, normalization, weights, or quantization path, so it cannot safely query, repair, or migrate that index.

Acceptable fixes remain:

1. fail closed when all remote endpoints fail; or
2. persist the actual producer identity/fingerprint per vector and require a separate index/full re-embed when the producer changes.

A local fallback must not write into the remote producer's fingerprint, even as an explicit opt-in. The existing `CHANGES_REQUESTED` therefore remains in force, but the PR is now narrower and easier to review once the producer-identity issue is fixed.

epruseal:
Removed the local-GGUF-last-resort fallback entirely (ec1e6e6) instead of gating it behind an opt-in flag, per your point that consent doesn't resolve the underlying ambiguity — there's still no way to tell after the fact which vectors came from which producer.

`embed()`/`embedBatch()` now fail closed on total remote failure: they return `null` (and log the error), the same behavior a remote-only config with no local entry already had. Sequential fallback across multiple remote endpoints (the actual feature this PR adds) is untouched.

Also dropped the three tests that codified the old fallback behavior; the existing "return null on total remote failure" and "mismatched #model-id fragments throw" tests already cover the fail-closed contract and identity validation. Full suite: 921 passed, 0 failed, tsc clean.

PR #766, opened 2026-07-13
fix(store): make cleanupOrphanedVectors atomic across both vector tables

fa1k3:
## Summary

`cleanupOrphanedVectors()` runs its two DELETEs (`vectors_vec`, then `content_vectors`) as separate implicit transactions. An interruption between them — process crash, or `SQLITE_BUSY` from a concurrent writer hitting the second statement — desyncs the two tables: the embedding rows are already gone from `vectors_vec` while `content_vectors` still reports the chunks as embedded.

Since these rows are orphaned (no active document) and vector search post-filters on `documents.active = 1`, live search is unaffected right away. The failure is **latent**: qmd is content-addressable, so if the same content later returns and that hash is reactivated, the stale `content_vectors` rows make `getHashesNeedingEmbedding` treat it as already embedded — `qmd embed` skips it, and the document is silently unsearchable by vector with no orphan left for a future `cleanup` to catch.

## What changed

- The orphan count and both DELETEs now run inside a single `db.transaction()`, executed as `BEGIN IMMEDIATE`: the count reads before the DELETEs write, and upgrading a deferred read snapshot under a concurrent WAL writer fails with `SQLITE_BUSY_SNAPSHOT` instead of honoring the busy timeout. The shared `Database` interface type now reflects the `.immediate` variant both drivers already expose.
- The count shares the transaction so the returned number always matches what the DELETEs actually removed, even when another connection mutates `documents` concurrently.
- Nested callers stay safe: both better-sqlite3 and `bun:sqlite` implement nested transaction functions via savepoints (an immediate transaction invoked inside an outer transaction becomes a savepoint).
- Changelog entry under `[Unreleased]`.

## Verification

- New tests in `test/store.test.ts` (`cleanupOrphanedVectors atomicity`):
  - happy path: orphaned chunks removed from both tables, survivors intact, count correct
  - fault injection between the two DELETEs: the `vectors_vec` DELETE rolls back (this test fails against the previous code) and the connection stays usable for a plain retry
  - outer-transaction rollback also restores the cleanup's deletions
  - inner-savepoint isolation: a cleanup failure caught inside an outer transaction rolls back alone while the outer transaction's own write commits
- `bun run test:types` clean, `bun run build` passes
- Suites green under both runtimes (vitest via Node and `bun test`)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,16 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `cleanupOrphanedVectors` now runs its orphan count and both DELETEs in a
+  single immediate transaction. An interruption between the two DELETEs
+  (crash, `SQLITE_BUSY`) could desync `vectors_vec` from `content_vectors`,
+  leaving stale metadata rows that make a later reactivation of the same
+  content hash look already-embedded — so `qmd embed` skips it and the
+  document becomes silently unsearchable by vector, with no orphan left to
+  clean up.
+
 ## [2.6.3] - 2026-06-24
 
 ### Added
--- a/src/db.ts
+++ b/src/db.ts
@@ -129,7 +129,9 @@ export interface Database {
   exec(sql: string): void;
   prepare(sql: string): Statement;
   loadExtension(path: string): void;
-  transaction<T extends (...args: SQLiteValue[]) => unknown>(fn: T): T;
+  // Both drivers return the wrapped function with variant methods attached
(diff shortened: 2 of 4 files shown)


tobi:
Superseded by https://github.com/tobi/qmd/pull/856 — same atomic cleanup, rebased onto current main after #851 extracted countOrphanedVectors (CHANGELOG + that refactor were the conflicts here).

tobi:
Closed in favor of #856, which lands this fix on current main.

Issue #129, opened 2026-02-07
Bun segfaults on Windows after vsearch/query due to process.exit() during native addon unload


qwexs:
# Bun segfaults on Windows after `vsearch`/`query` due to `process.exit()` during native addon unload

## Description

On Windows, QMD crashes with a segmentation fault after every `vsearch` or `query` command. The search results are returned correctly, but Bun panics during process termination with exit code 3.

The root cause is `process.exit(0)` in `qmd.ts` (line ~2696) triggering synchronous unloading of `llama-addon.node` (the node-llama-cpp native addon). Bun crashes when attempting to deallocate Vulkan/GPU resources during forced DLL teardown.

## Environment

- **OS:** Windows 11 x64
- **Bun:** 1.3.8
- **QMD:** 1.0.0 (installed via `bun install -g https://github.com/tobi/qmd`)
- **node-llama-cpp:** 3.15.0
- **GPU:** AMD Radeon RX 6800 (RDNA2, Vulkan 1.4.315, AMD proprietary driver 26.1.1)

## Steps to reproduce

1. Install QMD on Windows:
   ```powershell
   bun install -g https://github.com/tobi/qmd
   ```

2. Create a collection and generate embeddings:
   ```powershell
   qmd collection add . --name test --mask "**/*.md"
   qmd embed
   ```

3. Run a vector search:
   ```powershell
   qmd vsearch "any query"
   ```

4. Observe: search results print correctly, then Bun crashes:
   ```
   panic(main thread): Segmentation fault at address 0x2355FBA6724
   oh no: Bun has crashed. This indicates a bug in Bun, not your code.
   ```

   The crash report URL points to `llama-addon.node` in the stack trace:
   ```
   https://bun.report/1.3.8/wa1b64edcbijEugggCqv1itBCgBllama-addon.node6juBCgBllama-addon.nodewruFCgBllama-addon.nodeyvwFCgBllama-addon.nodeqnhEu8rwlC4+ritByn1ktBwg4vC2k6zPqhvjrB__g544vCA2qjBoyzp3/C
   ```

5. The same crash occurs with `qmd query "any query"`.

6. `qmd search` (BM25 only, no GPU models) works fine — no crash.

## Root cause

In `src/qmd.ts`, the exit sequence is:

```typescript
await disposeDefaultLlamaCpp();
process.exit(0);
```

[...]

PR #767, opened 2026-07-13
fix(mcp): shut down gracefully on stdin EOF instead of orphaning to PID 1 (#751)

### LINKED ISSUE #751: qmd mcp (stdio) doesn't exit on stdin EOF — orphans to PID 1 and leaks RAM after parent dies
## Summary

When `qmd mcp` runs in **stdio** mode and its parent process (an MCP client like Claude Code / Codex) dies, the qmd process does **not** exit. It reparents to launchd/init (PPID=1) and lingers indefinitely, holding the open SQLite store + the native embedding model in memory. Across a day of client sessions these pile up — I was seeing multiple orphaned MCP servers eating hundreds of MB each on macOS.

## Root cause

`@modelcontextprotocol/sdk`'s `StdioServerTransport` registers stdin `data`/`error` listeners only — it never listens for `end`/`close`. So on parent-death EOF the transport's `onclose` never fires and nothing tears the process down.

The HTTP path already handles shutdown (SIGTERM / explicit exit), but the **stdio** path (`startMcpServer`, roughly `src/mcp/server.ts`) has no EOF/`onclose`/SIGTERM handling. Because qmd holds a native embedding binding + an open store, the event loop stays alive even with stdin gone, so it never exits naturally.

## Repro

1. Spawn `qmd mcp` from a parent process with piped stdin.
2. Kill the parent (or close its stdin).
3. Observe the qmd process survive, reparented to PID 1: `ps -o pid,ppid,rss,command -p <pid>`.

## Suggested fix

Exit when stdin closes, after `await server.connect(transport)`:

```js
process.stdin.on("close", () => {
  // tear down the store + embedding model first, then exit
  shutdown();        // close SQLite, release native model
  process.exit(0);
});
```

## Caveat — do the teardown cleanly

`process.exit()` during native-addon unload has bitten qmd before:
- #59 (SIGABRT on exit during Metal cleanup, macOS ARM64)
- #129 (Bun segfault on `process.exit()` during native addon unload)

So the stdio EOF handler should release the embedding model / store gracefully before exiting (or set `process.exitCode` and let the loop drain) to avoid re-triggering those exit-time crashes.

## Context

I patched the same `@modelcontextprotocol/sdk` stdin-EOF gap in two of my own MCP servers with a one-line `process.stdin.on("close", …)`; qmd is the remaining offender in my setup and the only one I can't patch since it's an installed package. Happy to send a PR if you'd like — just wanted to flag the exit-time-crash interaction first.

fa1k3:
## Summary

Fixes #751.

The SDK's `StdioServerTransport` subscribes to stdin `data`/`error` only and never notices `end`/`close`. When the parent MCP client dies, no teardown runs: whenever anything still references the event loop (a warm llama.cpp model's native handles being the common case), `qmd mcp` reparents to PID 1, leaks RAM, and keeps the SQLite index open. Multiple such orphans can pile up across client sessions.

## What changed

`startMcpServer()` now registers `registerStdioEofShutdown()` after `server.connect()`. On stdin EOF it treats the client as disconnected and tears down gracefully:

1. `server.close()` — stop the transport, no further requests are dispatched
2. **in-flight drain** — a new `InflightGate` counts running tool/resource handlers; shutdown waits (bounded, 5 s default) for them to settle so their store/llm dependencies are not torn down underneath them
3. `store.close()` — last, and it disposes this store's own `LlamaCpp` instance before closing the database. (The shutdown helper takes an optional `disposeLlm` step for callers that own a separate instance, but the MCP path deliberately does not pass it: `store.close()` already owns the disposal, and defaulting to the global `disposeDefaultLlamaCpp` would risk tearing down an unrelated instance in an embedded process.)
4. `process.exitCode = 0` (or 1 if a step failed; an earlier nonzero code is preserved, not masked) and let the loop drain

Design decisions, mirroring existing house patterns:

- **No `process.exit()`.** Setting `process.exitCode` keeps `beforeExit` alive so node-llama-cpp's auto-dispose runs before libc's static destructors — forced exits during native-addon unload have crashed before (#59, #129). Same rationale as `finishSuccessfulCliCommand` on the CLI side.
- **EOF means disconnect.** No response-delivery promise is made — with stdin at EOF there is nobody left to read one. Handlers that are already running get the bounded drain window above; the SDK's close semantics still signal cancellation. Because `server.close()` runs before the drain, the transport stops dispatching new requests first, so the gate only ever waits on already-dispatched work (its count is not a closed admission barrier, and does not need to be under this ordering).
- **Idempotent shared-promise shutdown** (like the HTTP transport's `stop()`): `end`, `close`, an already-ended stdin at registration time, and manual invocations all collapse into one run; listeners are removed on entry.
- **Failure-tolerant chain:** a failing step is logged (`QMD Warning: … continuing shutdown.`) and does not stop the remaining steps.
- **stderr breadcrumb** `Shutting down (stdin closed)...`, matching the HTTP transport's SIGTERM/SIGINT messages, and giving tests an observable signal that the EOF path (not a coincidental event-loop drain) produced the exit. All logging is best-effort: the parent's death may have closed stderr too, so writes can never take the teardown down with them.
- **No forced-exit watchdog.** A timer that force-exits mid-cleanup would re-open exactly the #59/#129 crash class. In testing (Linux, warm embedding model), disposal + natural drain exits within ~100 ms of EOF; if a reproducible post-dispose hang shows up on some platform, a watchdog armed only *after* the teardown chain completes would be the safe follow-up.

Related: modelcontextprotocol/typescript-sdk#2003 (open, not yet merged) proposes fixing the same gap at the transport layer. This change stays correct alongside it once it lands (the teardown is idempotent), and qmd needs its own store/llm cleanup either way.

Note: the changelog entry lands in the same `[Unreleased] › Fixed` block as #766 (the atomic-cleanup PR), so a trivial changelog conflict is expected if both merge — happy to rebase whichever lands second.

## Verification

- New `test/mcp-stdio-lifecycle.test.ts`:
  - unit tests drive `registerStdioEofShutdown` via injected stdin (mirroring the DI style of the `finishSuccessfulCliCommand` tests): teardown order incl. the in-flight drain, `end`/`close`/manual idempotency, failure tolerance incl. all-steps-failing and a throwing stderr, drain-deadline miss, exit-code preservation, already-ended/destroyed stdin at registration
  - `createInflightGate` unit tests: immediate idle, settle-wait, missed deadline, rejecting handler releases the gate
  - end-to-end: spawns the real server with piped stdio, completes an `initialize` round-trip, closes stdin, and asserts a prompt code-0 exit **through the EOF shutdown path** (stderr breadcrumb; without it a modelless server can exit 0 by coincidence)
- Manually verified with a warm embedding model (Linux): the patched server logs the breadcrumb and exits code 0 within ~100 ms of stdin EOF. (I could not reproduce the indefinite hang itself on this Linux box — whether something keeps the loop alive after EOF is platform/timing dependent — but with this change the teardown no longer relies on the loop happening to drain.)
- `bun run test:types` clean, `bun run build` passes, suites green under vitest via Node and `bun test`
- Changelog entry under `[Unreleased]`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -23,6 +23,13 @@
 
 ### Fixed
 
+- `qmd mcp` (stdio) now shuts down gracefully when stdin reaches EOF instead
+  of orphaning to PID 1 when the parent MCP client dies (#751): the server
+  closes its transport, gives in-flight request handlers a bounded window to
+  settle, closes the store (which disposes its llama.cpp instance), and lets
+  the process drain via `process.exitCode` (no forced `process.exit()`, which
+  has caused exit-time native crashes before).
+
 - `qmd --version` no longer reports an unrelated repository's commit (#787).
   The commit was discovered at runtime with `git -C <installDir> rev-parse`,
   and `git -C` walks *up*, so any install nested inside another checkout
@@ -43,17 +50,20 @@
   `qmd pull` next to each download) as invalid GGUF models. The model-cache
   check now only inspects real `.gguf` files, so a sidecar that happens to
   sort before the blob no longer poisons the report. #812
+
 - `qmd mcp stop` and `qmd mcp --http --daemon` now verify that a pidfile PID still belongs to a qmd process before signalling it or refusing to start. Recycled PIDs (common after reboot) are treated as stale: the pidfile is unlinked instead of SIGTERM'ing an unrelated process or blocking daemon start with a false "Already running" error (#806).
 
 - `qmd collection add` now rejects missing paths and regular files before
   creating collection configuration or index state. The error reports both the
   received and resolved path so malformed shell arguments can be corrected.
+
 - Claude Code plugin: scope the plugin `source` to `./skills` so installs
   copy just the skills (~50 KB) instead of the entire repository. Previously a
   canonical install materialized ~230 MB / 9,000+ items into
   `~/.claude/plugins/cache/` — including a full npm dependency install
   triggered by the repo-root `package.json`. Both skills (`qmd` and `release`)
   still ship, unchanged. (#790)
+
 - Claude Code plugin: releases now bump the plugin version in
   `.claude-plugin/marketplace.json` in lockstep with `package.json`. The
   plugin cache is keyed on this version, and it had been stuck at `0.1.0`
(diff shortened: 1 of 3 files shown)


tobi:
Superseded by #836 after rebase onto main.

tobi:
Superseded by #836 after rebase onto main.

tobi:
Superseded by #837 (relanded onto current main; CHANGELOG conflict resolved). Closing in favor of the reland.

PR #777, opened 2026-07-19
fix: don't split UTF-16 surrogate pairs at chunk boundaries

junmo-kim:
## Problem

When a document being chunked for embedding has an astral-plane character (an emoji, for instance) sitting right at a chunk boundary, the chunker can cut the raw UTF-16 offset in the middle of that character's surrogate pair. The resulting chunk text then contains an unpaired surrogate, and some OpenAI-compatible remote embedding endpoints reject the JSON payload for it as invalid (`invalid string: surrogate U+DC00..U+DFFF must follow U+D800..U+DBFF`). Because the chunking is deterministic, the same chunk fails the same way on every retry, so the document's embedding gets stuck permanently.

This shows up in two places: `chunkDocumentWithBreakPoints`'s own boundary math, and `chunkDocumentByTokens`'s recursive re-splitting, which shrinks the char budget toward 1 for astral-plane-dense content with a high token/char ratio — small enough that a naive "give up" fallback for the boundary adjustment could reproduce the same split.

## Solution

Adds an `isSurrogatePairBoundary`/`adjustSurrogateBoundary` helper pair and applies it at the two points in `chunkDocumentWithBreakPoints` where a chunk boundary is finalized: the end of the current chunk (right before slicing) and the start of the next chunk (after subtracting the overlap). Both prefer nudging the cut back by one code unit to exclude the pair from this chunk; when that would stall the loop's forward progress (the char budget is too small to retreat), they advance by one code unit instead, including the whole pair rather than splitting it. Advancing is always safe — the pair's low surrogate is never the last character of the document when this fires, and the position is already known to be ahead of the previous chunk's start. This guarantees a pair is never split, regardless of how small the budget gets, and covers `chunkDocument`, `chunkDocumentAsync`, and `chunkDocumentByTokens`'s recursion for free, since they all route through this function.

As defense-in-depth, `chunkDocumentByTokens` also strips any unpaired surrogate immediately before each of its two result-emitting sites, via a small `stripUnpairedSurrogates` helper. This covers two paths the boundary fix above doesn't reach: its token-budget truncation fallback reconstructs text via the tokenizer's `detokenize()`, which can itself land mid-character if a single astral-plane character was encoded across multiple tokens; and a lone surrogate that was already malformed in the source document (an unrelated upstream encoding issue, not something chunking created) can land at an ordinary chunk boundary by coincidence, since chunking only adjusts boundaries that split a pair it can see. The helper checks all four ways a lone surrogate can sit at a boundary — leading lone low, leading lone high not followed by its low, trailing lone high, and trailing lone low not preceded by its high — distinguishing a genuine unpaired surrogate from the edge of a valid, complete pair.

## Tests

Regression tests in `test/store.test.ts`:
- Two tests for `chunkDocumentWithBreakPoints` construct content where an emoji sits exactly on the raw fallback boundary and on the overlap-derived boundary.
- Five tests for `chunkDocumentByTokens` (mocking the tokenizer, following the existing `setDefaultLlamaCpp` pattern used by the neighboring pathological-blob test) cover the recursive re-split path, the `detokenize()` truncation fallback, an already-malformed single-character document, and a lone surrogate landing at the leading/trailing edge of an ordinary (non-degenerate) chunk.

All seven fail on `main` and pass with this change (verified via revert-and-rerun, not just red-then-green ordering). The existing `chunkDocumentWithBreakPoints`/`chunkDocument` equivalence test and the full test suite (vitest + bun) remain green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `qmd embed` no longer splits a UTF-16 surrogate pair (emoji, etc.) across a
+  chunk boundary. A chunk ending or starting mid-pair produced an unpaired
+  surrogate in the chunk text, which some remote embedding APIs reject as
+  invalid JSON — permanently failing that chunk on every retry, since the
+  boundary calculation is deterministic.
+
 ## [2.6.3] - 2026-06-24
 
 ### Added
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,10 @@
   chunk boundary. A chunk ending or starting mid-pair produced an unpaired
   surrogate in the chunk text, which some remote embedding APIs reject as
   invalid JSON — permanently failing that chunk on every retry, since the
-  boundary calculation is deterministic.
+  boundary calculation is deterministic. This covers both the character-based
+  chunker and `chunkDocumentByTokens`'s recursive re-splitting for
+  astral-plane-dense content, which could previously drive the char budget
+  low enough to reproduce the same split.
 
 ## [2.6.3] - 2026-06-24
 
--- a/src/store.ts
+++ b/src/store.ts
@@ -282,17 +282,27 @@ function isSurrogatePairBoundary(content: string, pos: number): boolean {
 }
 
 /**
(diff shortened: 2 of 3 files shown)


tobi:
Landed on main via #855 (rebased onto current main; CHANGELOG-only conflict).

PR #782, opened 2026-07-23
fix: surface the real error when rerank context creation fails

sv789:
`ensureRerankContexts()` (`src/llm.ts`) catches context-creation failures with bare `catch {}`, so the actual node-llama-cpp error — typically an out-of-VRAM condition — is discarded and every failure reads as an uninformative `Failed to create any rerank context`. An OOM and a corrupt model file are indistinguishable from the output.

The fallback retry is also dead code: its comment says *"Flash attention might not be supported — retry without it"*, but `LlamaRankingContextOptions` has no `flashAttention` field (node-llama-cpp 3.18.1 never passes one when creating ranking contexts), so the retry repeats byte-identical options and can never recover.

This PR:
- includes the underlying message in the thrown error and chains the original via `{ cause: error }`;
- removes the impossible retry and corrects the stale tuning comment that budgeted VRAM for flash attention.

Verified on Windows 11, RX 570 8 GB (Vulkan), with the card under memory pressure from another process:

**Before:** `Error: Failed to create any rerank context`

**After:**
```
Error: Failed to create any rerank context: A context size of 40960 is too large for the available VRAM
  [cause]: InsufficientMemoryError: A context size of 40960 is too large for the available VRAM
```

### DIFF
--- a/src/llm.ts
+++ b/src/llm.ts
@@ -1170,10 +1170,8 @@ export class LlamaCpp implements LLM {
    * Load rerank contexts (lazy). Creates multiple contexts for parallel ranking.
    * Each context has its own sequence, so they can evaluate independently.
    *
-   * Tuning choices:
-   * - contextSize 1024: reranking chunks are ~800 tokens max, 1024 is plenty
-   * - flashAttention: ~20% less VRAM per context (568 vs 711 MB)
-   * - Combined: drops from 11.6 GB (auto, no flash) to 568 MB per context (20×)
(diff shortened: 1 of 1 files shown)


tobi:
Superseded by #866, which rebases this onto current main (warn-and-skip rather than throw, but the warning now includes the underlying node-llama-cpp error and the dead identical flash-attention retry is gone).

tobi:
Closing in favor of #866.

PR #784, opened 2026-07-24
fix(store): stop legacy path migration from evicting live files (#717)

### LINKED ISSUE #717: Lossy path normalization breaks QMD_EDITOR_URI for files with spaces
This is a repost of #45, which was closed requesting a reproduction on the latest version.

QMD normalizes filenames with hyphens, so files containing spaces or underscores won't work properly with QMD_EDITOR_URI.

### Reproduction steps (using qmd 2.5.3 (179d450))
```bash
mkdir /tmp/qmd-lossy
cd /tmp/qmd-lossy
echo "file with spaces" > "my file.md"
echo "file with hyphens" > my-file.md
qmd collection add /tmp/qmd-lossy --name qmd-lossy
qmd ls qmd-lossy
# Only one file is shown

export QMD_EDITOR_URI="obsidian://open?file={path}"
# The path substituted in links always uses hyphens even if the original path uses spaces
```

danmackinlay:
Fixes #717 — the parts of it that are still reproducible on `main` (v2.6.3).

Index-time slugging is already gone (070147d stores literal filesystem paths), but two lossy-path bugs survived it, and both still cost snake_case corpora documents today.

## 1. The legacy-path migration evicts live files

`findOrMigrateLegacyDocument()` reconstructs the path a pre-2.6 index would have stored (`handelize("2026_06_16.md")` → `"2026-06-16.md"`) and adopts the row it finds. That inverse is lossy — `a b.md`, `a_b.md` and `a-b.md` all handalize to `a-b.md` — and the lookup never checked whether the row it matched belongs to a file that still exists. So it renames a live file's row onto the new document, and the live file disappears from the index.

On `main` today:

```sh
mkdir -p /tmp/qmd-717 && cd /tmp/qmd-717
echo hyphen     > 2026-06-16.md
echo underscore > 2026_06_16.md
qmd collection add /tmp/qmd-717 --name t717
qmd ls t717
# qmd://t717/2026_06_16.md      <- only one row; 2026-06-16.md is gone
```

The incremental form is worse, because it removes a document that was already indexed and searchable: index `2026-06-16.md` alone, then add `2026_06_16.md` and run `qmd update` — the hyphenated file is dropped. This is the OP's "Only one file is shown", and @manpreetmaso's Logseq journals (`2026_06_16.md`) are the worst case since every filename in that corpus has the collision.

Both legacy lookups now skip a candidate row whose path is still owned by a file in the current scan, so only genuinely stale rows are adopted. Legitimate migration of a pre-2.6 index is unaffected (covered by the existing migration test, plus a new one where the legacy slug collides with a live file).

**`handelize()` is deliberately left alone.** It is no longer a display or index function — its only caller is this migration, where its job is to reproduce the old path format byte for byte. Making it preserve `_` would mean pre-2.6 snake_case indexes stop matching and stop migrating in place — precisely the users this issue is about. Added a doc comment saying so, since "why does this still slugify underscores?" is an obvious next question.

## 2. `_` in a path is a SQL `LIKE` wildcard

Same report, different mechanism, and it returns the *wrong document* rather than none:

```sh
qmd get 2026_06_16.md    # printed the contents of 2026-06-16.md
qmd ls notes/2026_06     # listed notes/2026-06-… too
```

`qmd get`, `qmd multi-get` and `qmd ls <prefix>` interpolate the user's path straight into a `LIKE` pattern, where `_` matches any character and `%` matches anything. Added `escapeLikePattern()` + an `ESCAPE` clause at the four call sites. I found this because the regression test for (1) failed on it.

## Tests

`test/path-fidelity.test.ts` gains a "separator collisions (#717)" block: fresh index keeps all four of `2026-06-16.md` / `2026_06_16.md` / `my file.md` / `my-file.md`; adding an underscored twin to an indexed collection does not evict the hyphenated file (and each path still resolves to its own content); `_` is matched literally by `multi-get` and `ls`; and a legacy row still migrates when a live file shares its slug. Plus a unit test for `escapeLikePattern()`.

Full suite passes (952 passed / 7 skipped). Two pre-existing failures on my machine are unrelated to this change and fail identically on a clean `main`: a batch of `cli.test.ts` cases that assert `stderr === ""` and pick up Node's `DEP0205` warning under this Node version, and `mcp.test.ts > MCP HTTP Transport` (`SqliteError: no such column: T.name`). Both go away with `NODE_OPTIONS=--no-deprecation` / are environmental.

## Not addressed here

`--full-path` still falls back to the `qmd://` URI when a result can't be resolved on disk, and drops the docid while doing it (`get`/`multi-get` keep it). With literal paths stored, that fallback now only fires for files that genuinely no longer exist, so it reads as a stale-index signal rather than a normalization bug — happy to follow up separately if you want it to warn or keep the docid.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Files whose names differ only in the characters the legacy slug collapsed to
+  `-` (spaces, underscores) no longer evict each other from the index (#717).
+  The legacy-path migration reconstructs the slug a pre-2.6 index would have
+  stored (`2026_06_16.md` → `2026-06-16.md`) and adopts the matching row — but
+  it did so even when that row belonged to a live file, renaming it onto the new
+  document. Indexing `2026-06-16.md` and `2026_06_16.md` in the same collection
+  silently dropped one of them; a `qmd update` after adding an underscored twin
+  removed the hyphenated file from the index entirely. The migration now skips
+  any row whose path is still owned by a file in the current scan, so it only
+  ever adopts genuinely stale rows. `handelize()` is unchanged — it exists only
+  to reconstruct pre-2.6 paths, and altering its output would strand snake_case
+  corpora on the old index format.
+- `qmd get`, `qmd multi-get` and `qmd ls <prefix>` now match `_` and `%` in
+  paths literally instead of as SQL `LIKE` wildcards (#717). `qmd get
+  2026_06_16.md` could return the contents of a sibling `2026-06-16.md`, and
+  `qmd ls notes/2026_06` listed files it did not match.
+
 ## [2.6.3] - 2026-06-24
 
 ### Added
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -57,7 +57,7 @@ import {
   vacuumDatabase,
   getCollectionsWithoutContext,
   getTopLevelPathsWithoutContext,
-  handelize,
+  escapeLikePattern,
   hybridQuery,
   vectorSearchQuery,
   structuredSearch,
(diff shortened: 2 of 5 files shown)


danmackinlay:
Superseded by bf4b1af1 (#869), which fixes the same migration-eviction bug from #717, and 8bbb5b9b (#868), which covers the multi-get path resolution. Closing; no action needed.

PR #786, opened 2026-07-24
fix(cli): stop --full-path from degrading silently (#785)

### LINKED ISSUE #785: --full-path silently drops the docid and falls back to qmd:// when a result can't be resolved on disk
`--full-path` is documented as "Show on-disk paths instead of `qmd://` + docid". When a result can't be resolved on disk it falls back to the `qmd://` URI — which is reasonable — but in `search`/`query` the fallback is unannounced *and* drops the docid, so the row ends up with neither a usable filesystem path nor an identifier field. Nothing in the output distinguishes it from a resolved row.

Since 070147d (literal paths) this no longer fires for normalization reasons; it fires when the file genuinely moved or was deleted since the last index. That is useful information — a stale-index signal — and right now it is silently swallowed.

### Reproduction (qmd 2.6.3)

```bash
mkdir -p /tmp/qmd-fp/corpus && cd /tmp/qmd-fp/corpus
echo "alpha" > alpha.md
echo "beta"  > beta.md
qmd collection add /tmp/qmd-fp/corpus --name stale
mv beta.md /tmp/qmd-fp/beta-moved.md   # index now has a stale row
qmd search beta --full-path --json
```

```json
[
  { "score": 0, "file": "./alpha.md",          "line": 1, "title": "Alpha" },
  { "score": 0, "file": "qmd://stale/beta.md", "line": 1, "title": "Beta"  }
]
```

The second row has no `docid` field and a `file` value that is not a path. Same for `--format csv` (the `--full-path` header is `score,file,…` with no docid column) and `--files`. Nothing on stderr, exit code 0.

### `multi-get` already gets this right

```
./alpha.md
#4cbea2 qmd://stale/beta.md
```

`multiGet()` decides per row — `docidOf = (fullPath && r.fsPath) ? undefined : r.docid` — so an unresolved row keeps its docid and stays addressable. `qmd get --full-path` likewise falls back to `qmd://… #docid`. Only `outputResults()` (search/query) drops the docid based on the flag alone, regardless of whether that particular row resolved.

### Suggested fix

1. Make search/query decide per row, like `multi-get`: resolved → fs path, no docid; unresolved → `qmd://` URI **with** the docid. (For `--format csv` that means always emitting the docid column under `--full-path`, empty when the row resolved — which is what `multi-get`'s CSV already does.)
2. Say something on stderr when any row falls back, e.g. `warning: 1 of 2 results could not be resolved on disk (moved or deleted since indexing) — run 'qmd update'`. stdout stays machine-clean; the human gets the stale-index signal.

Splitting this out of #717 since it is a separate mechanism — noted at the bottom of #784, which fixes the normalization side. Happy to send a PR.

danmackinlay:
Fixes #785. Independent of #784 — branched from `main`, no overlap beyond both adding a `[Unreleased]` changelog block.

`--full-path` exists to emit openable filesystem paths. Falling back to the `qmd://` URI when a result can't be resolved is the right call, but `search`/`query` did it unannounced **and** dropped the docid, so the row ended up with neither a usable path nor an identifier field — and nothing distinguished it from a resolved row.

Since 070147d (literal paths) this no longer fires for normalization reasons. It fires when the file moved or was deleted since the last index, which is useful information that was being swallowed.

### Before

```console
$ qmd search beta --full-path --json      # beta.md was moved after indexing
[
  { "score": 0, "file": "./alpha.md",          "title": "Alpha" },
  { "score": 0, "file": "qmd://stale/beta.md", "title": "Beta"  }
]
$ echo $?
0
```

No `docid` on the second row, nothing on stderr. `qmd get --full-path` and `multi-get --full-path` already kept the docid — only `outputResults()` keyed the decision off the flag rather than off the row.

### After

```console
$ qmd search beta --full-path --json
[
  { "score": 0, "file": "./alpha.md",          "title": "Alpha" },
  { "docid": "#4cbea2", "score": 0, "file": "qmd://stale/beta.md", "title": "Beta" }
]
warning: --full-path could not resolve 1 of 2 results on disk (moved or deleted since indexing); showing qmd:// + docid instead. Run 'qmd update' to refresh the index.
```

The warning is on stderr, so stdout stays machine-readable and pipelines are unaffected.

## Changes

1. **Per-row decision in `search`/`query`**, matching `multiGet()`: resolved → on-disk path, no docid; unresolved → `qmd://` URI *with* docid. Applies to `cli`, `json`, `md`, `xml`, `csv`. Rows are resolved up front so the count is known before anything prints.
2. **stderr notice** from `search`, `query`, `get` and `multi-get` when any row falls back, pointing at `qmd update`.
3. **`search --format csv` always emits the `docid` column** (empty when the row resolved), so columns stay positional. The `--full-path` header used to drop it — which also disagreed with `printEmptySearchResults()`, whose CSV header always carries `docid`, so an empty result set and a non-empty one printed different headers for the same command.

### Two judgement calls worth a look

- **`--files` is left alone.** Its `--full-path` shape stays `score,path`, so `cut -d, -f2 | xargs` keeps working; adding a leading docid field only for unresolved rows would break positional parsing, and adding it unconditionally would break every existing pipeline. Unresolved rows there show the `qmd://` URI and are covered by the stderr notice. Happy to change it if you'd rather have consistency with CSV.
- **Exit code stays 0.** A moved file is a stale index, not a failed search. The notice is the signal.

## Tests

New `test/full-path-fallback.test.ts` (7 cases) over a corpus where one file is moved out from under the index: docid retention in `--json`, the CSV column layout, docid in the default CLI format, `get` and `multi-get` warnings, no warning when everything resolves, and no behavior change without `--full-path`. Existing `--full-path` tests in `cli.test.ts` (which assert resolved rows still drop the docid) pass unchanged.

Full suite: 954 passed / 7 skipped. `mcp.test.ts > MCP HTTP Transport` fails on my machine (`SqliteError: no such column: T.name`) identically on a clean `main` — environmental, unrelated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## [Unreleased]
 
+### Changed
+
+- `--full-path` no longer degrades silently when a result cannot be resolved on
+  disk (#785). A fallback there means the file moved or was deleted since the
+  last index, so `search`, `query`, `get` and `multi-get` now print a notice to
+  stderr naming how many results fell back and suggesting `qmd update`; stdout
+  stays machine-readable.
+- `search`/`query` now decide per result whether to show the docid under
+  `--full-path`, matching `multi-get` and `get`: a result that resolved shows
+  its on-disk path and no docid, one that did not keeps its `qmd://` URI *and*
+  its docid, so it is still addressable. Previously the docid was dropped for
+  every row whenever the flag was set, leaving unresolved rows with neither a
+  usable path nor an identifier.
+- `search --format csv` always emits the `docid` column, empty for rows that
+  resolved to an on-disk path. Under `--full-path` the header previously
+  dropped the column entirely — which also disagreed with the empty-result
+  header, always printed with `docid`. Column positions are now stable across
+  runs and formats.
+
 ### Fixed
 
 - `qmd --version` no longer reports an unrelated repository's commit (#787).
--- a/README.md
+++ b/README.md
@@ -790,6 +790,9 @@ and `deep-search` (→ `query`).
 --no-rerank        # Skip LLM reranking (RRF scores only; faster on CPU)
 -C, --candidate-limit <n>  # Max candidates to rerank (default: 40)
 --full-path        # Emit on-disk filesystem paths instead of qmd:// URIs
+                   # (a result whose file has moved or been deleted since
+                   #  indexing keeps its qmd:// URI + docid, and a notice is
+                   #  printed to stderr — run `qmd update` to refresh)
 
 # Output formats (for search and multi-get)
 --format <kind>    # cli (default) | json | csv | md | xml | files
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -970,6 +970,26 @@ function renderFullPath(absolutePath: string, cwd: string = process.cwd()): stri
   return real;
 }
 
+/**
+ * Report rows that `--full-path` could not turn into an on-disk path.
+ *
+ * The flag's whole job is producing openable paths, so falling back to a
+ * `qmd://` URI is worth saying out loud: it means the file moved or was
+ * deleted since the last index, not that the path was normalized away. The
+ * notice goes to stderr so stdout stays machine-readable.
+ */
+function warnUnresolvedFullPaths(unresolved: number, total: number): void {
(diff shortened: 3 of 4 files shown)


tobi:
Superseded by the reland PR after rebase onto main.

tobi:
Superseded by #835 after rebase onto main.

PR #788, opened 2026-07-24
fix(cli): stamp the build's commit instead of guessing at runtime (#787)

### LINKED ISSUE #787: `qmd --version` reports the enclosing directory's git HEAD, not qmd's
## Problem

`qmd --version` appends a commit hash obtained by running `git rev-parse` from its own install directory. `git -C <dir>` walks *up* the tree, so for any install that happens to sit inside some other repository, the reported hash is that repository's HEAD — nothing to do with qmd.

The most common way to hit this is a global npm install under a git-managed prefix. On macOS/Homebrew, `/opt/homebrew` is itself a git checkout:

```
$ qmd --version
qmd 2.5.3 (b48c7994b5)
$ git -C /opt/homebrew rev-parse --short=10 HEAD
b48c7994b5          # Homebrew's HEAD, reported as qmd's commit
```

Same package, two locations, two different "commits":

```sh
for r in alpha beta; do
  mkdir -p /tmp/vers/$r && cd /tmp/vers/$r
  git init -q . && git commit -q --allow-empty -m "unrelated repo $r"
  cp -R "$(npm root -g)/@tobilu/qmd" /tmp/vers/$r/qmd
  echo "$r HEAD=$(git rev-parse --short HEAD)  ->  $(/tmp/vers/$r/qmd/bin/qmd --version)"
done
```

```
alpha HEAD=8e68656  ->  qmd 2.6.3 (8e68656)
beta  HEAD=ba96930  ->  qmd 2.6.3 (ba96930)
```

Identical bytes, empty unrelated repos, and the version string faithfully reports each one. Copy the same directory somewhere with no enclosing repo and the hash disappears entirely (`qmd 2.6.3`), which is the intended fallback.

## Why it matters

Version strings in bug reports stop being evidence. #717 alone carries `2.5.3 (179d450)`, `2.5.3 (20ce57618d)` and `2.5.3 (b48c7994b5)` from three reporters — as far as I can tell, all the same published tarball, differing only in where each person's npm prefix happened to live. Anyone triaging on those hashes is chasing noise, and "works on `main`, reported on 2.5.3" is harder to establish than it should be.

It bit me directly: after installing a build of `main` over a 2.5.3 install, `--version` reported the *same* hash as before, which looks exactly like an install that silently didn't take.

## Root cause

`src/cli/qmd.ts` (`showVersion`, currently around L3967):

```ts
const scriptDir = dirname(fileURLToPath(import.meta.url));
...
commit = execSync(`git -C ${scriptDir} rev-parse --short HEAD`, ...);
```

The `catch` comment says "Not a git repo or git not available", which is the right intent — the problem is that `git -C` treats *any* ancestor repo as a hit, so the guard never fires where it should.

Two smaller things in the same line:

- `${scriptDir}` is interpolated into a shell string unquoted, so an install path containing a space makes git fail and the commit is dropped silently. Reproducible by copying the package into `.../a dir/qmd` inside a git repo: the hash vanishes, right answer by accident.
- It's a synchronous subprocess on a path where `--version` is expected to be instant.

## Suggested fix

Stamp the commit at build time rather than discovering it at runtime — a published tarball has no git history of its own, so runtime discovery can only ever find someone else's. `scripts/build.mjs` already runs where the real hash is knowable; injecting it (`__QMD_COMMIT__`, falling back to empty for source runs) makes the string correct for exactly the builds people file bugs from.

If you'd rather keep the runtime lookup for source checkouts, confirming the repo actually *is* qmd's before trusting it would do:

```ts
const top = execSync("git rev-parse --show-toplevel", { cwd: scriptDir, ... }).trim();
// only use the hash if `top` is the package root (e.g. top/package.json has name @tobilu/qmd)
```

Passing the directory via `cwd` instead of interpolating `-C ${scriptDir}` also disposes of the spaces bug.

Happy to send a PR for the build-time stamping if that's the direction you prefer.

## Environment

qmd 2.5.3 from npm and 2.6.3 built from `main` (both reproduce), Node 26, macOS 26.5, npm prefix `/opt/homebrew`.

danmackinlay:
Fixes #787. Independent of #784 and #786 — branched from `main`, the only overlap being a `[Unreleased]` changelog block.

`showVersion()` discovered the commit at runtime with `git -C <scriptDir> rev-parse --short HEAD`. `git -C` walks *up*, so any install nested inside another repository reports that repository's HEAD. A published tarball has no git history of its own, so runtime discovery can only ever find someone else's.

## Before / after, in the issue's own layout

A package copied into an unrelated repo (HEAD `96b9fc1`), with qmd itself built at `e428df7`:

```console
$ git -C <copy>/dist/cli rev-parse --short HEAD     # what showVersion() ran
96b9fc1                                             # the unrelated repo, reported as qmd's

$ <copy>/dist/cli/qmd.js --version                  # after, stamp removed
qmd 2.6.3                                           # no commit beats a false one

$ <copy>/dist/cli/qmd.js --version                  # after, as built
qmd 2.6.3 (e428df7-dirty)                           # the commit it was actually built from
```

## What changed

**`scripts/build.mjs` stamps the commit.** It already post-processes `dist/cli/qmd.js` to prepend the shebang, so the seam existed; it now also writes `dist/cli/build-info.json`. Uses `spawnSync` directly rather than the file's `run()` helper, which `process.exit`s on failure — no git, no repo, or a source tarball must degrade to an unstamped build, not a failed one. Stamping is itself guarded by the toplevel check, so building from inside an unrelated checkout can't stamp that checkout's commit. A build from a modified tree is marked `-dirty`, which answers "did my install actually take?" directly — the symptom that sent me looking.

**`src/cli/version.ts` (new)** — `readStampedCommit` / `gitCommitForCheckout` / `resolveCommit`. The stamp wins; a source checkout falls back to git, but only after confirming `git rev-parse --show-toplevel` is this package's root. Anything else reports no commit rather than a wrong one. It reads the stamp from beside the *running* script (`dist/cli/`) rather than from the package's `dist/`, so a source run can't pick up a stale stamp left by an earlier build.

**The lookup no longer goes through a shell.** `execFileSync` + `cwd` instead of interpolating `-C ${scriptDir}` into a shell string, so an install path containing a space stops silently dropping the commit. And `--version` from a build now spawns nothing at all.

Output format is unchanged: `qmd 2.6.3`, `qmd 2.6.3 (abc1234)`, or `qmd 2.6.3 (abc1234-dirty)`.

## Tests

- **`test/version.test.ts`** (new, ~0.5s, no build) — the #787 scenario as a unit test: a package nested inside an unrelated temp repo yields `""` and specifically *not* the outer HEAD; a package root that is the repo yields the real short HEAD; a dirty tree is marked; a path **containing a space** works; no repo at all yields `""`; the stamp is read, tolerates malformed/absent JSON, and beats the enclosing checkout.
- **`test/esm-ambiguous-module.test.ts`** — this was the only test executing the version path (it built, then asserted `toContain("qmd ")`). Moved its build into `beforeAll` and added a second case asserting the stamp file exists, that the printed string matches the exact format, and that the stamped commit is this repo's HEAD. No extra build; CI already runs this file on every PR.

Full suite: 959 passed / 7 skipped. `mcp.test.ts > MCP HTTP Transport` fails on my machine (`SqliteError: no such column: T.name`) identically on a clean `main` — environmental.

## Deliberately not included

- **MCP handshake version** (`src/mcp/server.ts:89-96`) has its own package.json reader and reports no commit. Wiring it to the same helper is a two-line change, but it alters a field MCP clients display, so it seemed worth asking rather than assuming.
- **`/health`** carries no version at all — left alone.
- **`flake.nix`** runs `src/` under bun and never calls `scripts/build.mjs`, and the store path has no `.git`, so nix builds report no commit. That's strictly better than today's wrong one, but if you want nix builds identifiable, the flake needs its own stamping step — happy to follow up.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,23 @@
 ## [Unreleased]
 
 ### Fixed
+
+- `qmd --version` no longer reports an unrelated repository's commit (#787).
+  The commit was discovered at runtime with `git -C <installDir> rev-parse`,
+  and `git -C` walks *up*, so any install nested inside another checkout
+  reported that checkout's HEAD — a global npm install under a git-managed
+  prefix such as Homebrew's `/opt/homebrew` claimed Homebrew's commit as
+  qmd's. Identical tarballs reported different "commits" depending only on
+  where they were installed, which is why one issue can collect three distinct
+  hashes for the same published build. `scripts/build.mjs` now stamps the
+  commit it built from into `dist/cli/build-info.json` (suffixed `-dirty` when
+  built from a modified tree), and the runtime prefers that. Source checkouts
+  still resolve their own HEAD, but only after confirming the enclosing
+  repository is qmd's; anything else reports no commit rather than a
+  misleading one. The lookup also no longer interpolates the install path into
+  a shell string, so a path containing a space stops silently dropping the
+  commit, and `--version` from a build runs no subprocess at all.
+
 - `qmd doctor` no longer false-positives `.etag` HTTP sidecars (written by
   `qmd pull` next to each download) as invalid GGUF models. The model-cache
   check now only inspects real `.gguf` files, so a sidecar that happens to
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -1,6 +1,6 @@
 #!/usr/bin/env node
 import { spawnSync } from "node:child_process";
-import { chmodSync, readFileSync, renameSync, writeFileSync } from "node:fs";
+import { chmodSync, readFileSync, realpathSync, renameSync, writeFileSync } from "node:fs";
 import { join } from "node:path";
 import { fileURLToPath } from "node:url";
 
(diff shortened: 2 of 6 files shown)


tobi:
Superseded by #834 after rebase onto main.

tobi:
Superseded by https://github.com/tobi/qmd/pull/833 (reland onto current main; original was conflicted). Closing in favor of the merged reland.

Issue #790, opened 2026-07-25
Claude Code plugin: ships the whole repo for a 24 KB skill


rymalia:
TL/DR: the plugin's `source` is the repo root (`"./"`), so every plugin install **installs** the entire qmd repository and node modules into `~/.claude/plugins/cache/` — when the plugin machinery only reads three files.

## What the plugin actually uses

- `skills/qmd/SKILL.md`
- `skills/qmd/references/mcp-setup.md`
- the manifest metadata

~24 KB

### Adding the marketplace clones the entire repo:
```bash
claude plugin marketplace add tobi/qmd
```
~11 MB, 179 items into `~/.claude/plugins/marketplaces/qmd/`

### Installing the plugin appears to also run a Node dependency install too:
```bash
claude plugin install qmd@qmd
```
The resulting folder in `~/.claude/plugins/cache/qmd/qmd/0.1.0/` is **230 MB with 9,000+ items**, including the `node_modules` folder. No one needs all that just for the plugin.

## Verified fix (tested locally)

Edit `marketplace.json`:
```json
"source": "./skills/qmd",
"skills": ["./"]
```

Install snapshot drops 230 MB → 24 KB, the skill arrives byte-identical (md5-checked), and `claude plugin details` parses the component inventory cleanly (Skills: 1). The declared `mcpServers` entry is unaffected — `command: "qmd"` resolves via PATH, not repo-relative.

## Note on the release skill

Scoping to `./skills/qmd` also stops shipping the `release` skill (`/qmd:release`) to plugin users. If shipping it to everyone is intentional, `"source": "./skills"` with `"skills": ["./qmd", "./release"]` keeps it while still trimming the payload to ~50 KB — either scoping captures nearly all of the win. Happy to PR this either way.

PR #794, opened 2026-07-26
fix: bump plugin version on every release so installed plugins receive updates

### LINKED ISSUE #789: bug: Claude Code plugin installs never receive updates — plugin version stuck at 0.1.0 since February
TL/DR: because the plugin version in `.claude-plugin/marketplace.json` hasn't changed since February, users are stuck with whatever version of the plugin they first installed — and never receive any of the updates shipped since. 😲

## Mechanism

Claude Code's plugin cache is keyed on the `version` field in `.claude-plugin/marketplace.json`. `claude plugin update` compares that version against the installed one — `0.1.0 == 0.1.0` — and reports "already at the latest version," regardless of how much the plugin's actual content has changed. There is no `--force` refresh; the only way is a manual uninstall + reinstall of the plugin, which users are not expected to do.

## Impact

Every installer is permanently pinned to the `skills/` snapshot from their install day:

- A February install still serves the February `skills/qmd/SKILL.md` (internal metadata `1.1.1`, "Quick Markdown Search", ~151 lines).
- The current skill is metadata `2.2.0`, ~295 lines — the structured `intent:`/`lex:`/`vec:`/`hyde:` query guidance, the `:from:count` line-range workflow, `--full-path`, the pitfalls section. None of it has reached anyone who installed prior to those changes being published.

## Verified repro

Long-standing install, Claude Code up to date:

```
$ ls ~/.claude/plugins/cache/qmd/qmd/
0.1.0                      # snapshot dated Feb 14; SKILL.md differs from repo HEAD
$ claude plugin update qmd@qmd
✔ qmd is already at the latest version (0.1.0).
```
As a controlled test (my marketplace registration points at a clone of this repo, so I could edit its manifest): changing **only** the version field — `0.1.0` → `0.1.1`, no content changes — and re-running the marketplace update and plugin reload `claude plugin update` delivered the current skill immediately. The marketplace's copy of the skill was already up to date the whole time; the version gate is the only thing holding updates back.

A second test: simply blank out the installed skill file at `~/.claude/plugins/cache/qmd/qmd/0.1.0/skills/qmd/SKILL.md`, run `claude plugin update`, and restart Claude Code. Invoking `/qmd:qmd` now serves the empty skill file — the updater left the edited cache untouched. While the marketplace version matches, `update` never looks at content: it won't deliver improvements, and it won't even repair a damaged install.

A third test, on a stock install (separate machine, plugin installed normally from `github:tobi/qmd`): edit the version field in the *downloaded* marketplace clone — `~/.claude/plugins/marketplaces/qmd/.claude-plugin/marketplace.json`, e.g. `0.1.0` → `0.1.9` — then run `/reload-plugins`. The reload does not re-pull from GitHub; it compares the clone's declared version against the installed one and, seeing it newer, refreshes the cache from the clone's (already-current) content. The stale skill updates immediately.

Together the tests bracket the mechanism exactly: content changes without a version change update nothing (not even a damaged cache gets repaired); a version change alone updates everything — on both local-clone and stock GitHub installs. Test 3 doubles as an interim workaround for anyone currently stuck, but it's a per-machine manual hack that gets overwritten on the next marketplace refresh — the real fix is a `marketplace.json` version bump in this repo, which updates every install through the normal path.

## Suggested fix

Stamp the plugin version in lockstep with `package.json` in `scripts/release.sh`, so cache invalidation is a free side effect of every release (no new manual step to remember), plus a one-time catch-up bump `0.1.0` → `2.6.3` so existing installs pick up the current skill on their next `claude plugin update`. Happy to PR this — the change is ~10 lines including the release-skill doc update.

rymalia:
Fixes #789. Related: #790 (same neglected manifest — the scoping fix there composes cleanly with this one).

## Problem

Claude Code's plugin cache is keyed on the plugin `version` declared in `.claude-plugin/marketplace.json` — and that version has been `0.1.0` since the plugin was introduced in February. `claude plugin update` compares version strings, sees `0.1.0 == 0.1.0`, and reports "already at the latest version" no matter how much the plugin's content has changed.

The consequence: **no installed plugin has ever received a skill update.** Every user is permanently pinned to whatever `skills/` contained on the day they first installed:

- A February installer is still served the February snapshot of `skills/qmd/SKILL.md` (internal metadata `1.1.1`, titled "Quick Markdown Search", ~151 lines).
- Today's `skills/qmd/SKILL.md` is at metadata `2.2.0`, ~295 lines — the structured `intent:`/`lex:`/`vec:`/`hyde:` query guidance, the `:from:count` line-range workflow, `--full-path`, the pitfalls section: none of it reaches previously-installed users.
- The only escape hatch is a manual `claude plugin uninstall` + `install` (there is no `--force` refresh), which nothing prompts users to do.

Verified on a long-standing install: the cached copy at `~/.claude/plugins/cache/qmd/qmd/0.1.0/skills/qmd/SKILL.md` was dated Feb 20 and differed from the repo's current file, while the marketplace clone was fully up to date — the version gate was the only thing holding the update back. Bumping the version and running `claude plugin update` immediately delivered the current skill.

## Fix

Two parts:

1. **`scripts/release.sh` now stamps the plugin version in lockstep with `package.json`** — cache invalidation becomes a free side effect of every release, with no new manual step to remember. `sed` rather than `jq` so the file's formatting is preserved (a `jq` round-trip reformats the inline arrays; the file has a single `"version"` key so the substitution is unambiguous). A `grep` guard fails the release if the stamp didn't land — a non-matching `sed` exits 0, which `set -e` alone would never catch.

2. **One-time catch-up: plugin version `0.1.0` → `2.6.3`** (the current package version), so existing installs pick up the current skill on their next `claude plugin update` instead of staying stranded.

Also documents the extra bump in the release skill (`skills/release/SKILL.md`) and adds a changelog entry under `[Unreleased]`.

After this, the plugin version tracks the package version (`2.6.4`, `2.7.0`, …) automatically on every `scripts/release.sh` run.

## Notes

- No MCP/runtime behavior change for the plugin; the release-skill docs are updated to mention the extra bump, and the version metadata and release tooling change.
- Existing installs still need one `claude plugin update qmd@qmd` (or reinstall) after this lands to cross the gap; from then on updates flow normally.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
 			"name": "qmd",
 			"source": "./skills",
 			"description": "Search and retrieve documents from local markdown files.",
-			"version": "0.1.0",
+			"version": "2.6.3",
 			"author": {
 				"name": "tobi",
 				"email": "tobi@lutke.com"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,15 @@
 
 ### Fixed
 
+- Claude Code plugin: releases now bump the plugin version in
+  `.claude-plugin/marketplace.json` in lockstep with `package.json`. The
+  plugin cache is keyed on this version, and it had been stuck at `0.1.0`
+  since February — so installed plugins never received skill updates
+  (users who installed in February are still being served that snapshot
+  today, despite the qmd skill nearly tripling in size since). Also bumps
+  the plugin to `2.6.3` as a one-time catch-up so existing installs pick
+  up the current skill on their next `claude plugin update`.
+
 - `qmd doctor` no longer false-positives `.etag` HTTP sidecars (written by
   `qmd pull` next to each download) as invalid GGUF models. The model-cache
   check now only inspects real `.gguf` files, so a sidecar that happens to
--- a/scripts/release.sh
+++ b/scripts/release.sh
@@ -108,7 +108,18 @@ awk '
 
 jq --arg v "$NEW" '.version = $v' package.json > package.json.tmp && mv package.json.tmp package.json
 
-git add package.json CHANGELOG.md
+# Keep the Claude Code plugin version in lockstep with the package version.
+# The plugin cache is keyed on this version: if it never changes, installed
+# plugins never pick up skill updates shipped in this release. (sed, not jq,
+# to preserve the file's formatting; the file has a single "version" key.)
+# The grep guard fails the release if the stamp didn't land (a non-matching
+# sed exits 0, which set -e would never catch).
+sed 's/"version": "[^"]*"/"version": "'"$NEW"'"/' \
+  .claude-plugin/marketplace.json > .claude-plugin/marketplace.json.tmp
+grep -qF "\"version\": \"$NEW\"" .claude-plugin/marketplace.json.tmp
+mv .claude-plugin/marketplace.json.tmp .claude-plugin/marketplace.json
+
+git add package.json CHANGELOG.md .claude-plugin/marketplace.json
 git commit -m "release: v$NEW"
 git tag -a "v$NEW" -m "v$NEW"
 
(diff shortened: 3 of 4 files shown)


tobi:
Superseded by #832 after rebase onto main.

tobi:
Superseded by #832 after rebase onto main.

tobi:
Superseded by https://github.com/tobi/qmd/pull/832

PR #801, opened 2026-07-29
fix: preserve case-sensitive document identities

jplew:
## Summary

- remove the implicit case-insensitive document lookup from legacy path migration
- preserve distinct active identities such as `README.md` and `readme.md` on case-sensitive filesystems
- retain exact-path lookup and the separate handalized-path migration
- add regression coverage for case-distinct documents and case-only renames

## Why

The previous `COLLATE NOCASE` fallback silently collapsed one document when a collection contained two paths that differed only by case. This is unsafe because the database cannot distinguish a genuine second document from a historical lowercased path without explicit migration provenance.

## Verification

- `bun test --preload ./src/test-preload.ts test/store.test.ts` — 223 passed, 0 failed
- `git diff --check` — passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Case-sensitive collections no longer collapse distinct document identities that
+  differ only by path casing. The implicit `COLLATE NOCASE` legacy migration was
+  unsafe for filesystems that contain both `README.md` and `readme.md`; case-only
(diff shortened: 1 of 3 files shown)


tobi:
Landed on main via #857 (rebased onto post-#856 main; CHANGELOG-only conflict).

jplew:
> Landed on main via #857 (rebased onto post-#856 main; CHANGELOG-only conflict).

arigatoo senpai 🙏

PR #802, opened 2026-07-29
feat(mcp): add update and embed maintenance tools

michabbb:
## Problem

MCP clients can already see a stale index via `status`, but they cannot do
anything about it. Re-indexing and embedding require leaving the MCP connection
and running `qmd update` / `qmd embed` in a shell — impossible in environments
without shell access. The dynamic server instructions even pointed at
`qmd embed`, a command the connected client has no way to call.

## What this adds

Two tools, `update` and `embed`, registered in the shared server factory so
stdio and Streamable HTTP expose the same implementation. Both call existing
store methods only — no CLI subprocess, no shell command, no source file is
modified, and configured collection update commands are never executed.

**`update`** syncs all configured collections or an explicit non-empty subset
and returns `collections`, `indexed`, `updated`, `unchanged`, `removed` and
`needsEmbedding`. Unknown names and an explicitly empty list are rejected
before the store is called, so a typo cannot look like a successful no-op.

**`embed`** supports scope, `force`, chunk strategy, batch bounds and a runtime
limit (default 30 minutes, `0` disables it, values above the safe Node timer
are rejected rather than silently overflowing to an immediate abort). The model
is never a client input.

## Safety boundaries

- Progress is emitted as MCP notifications only when the caller supplies a
  progress token. Delivery is best-effort and never changes the tool result.
- A per-process, per-store fail-fast lock permits one maintenance writer while
  read tools stay available; every exit path releases it. Separate processes
  keep relying on SQLite/WAL/busy-timeout.
- Cancellation is checked before the run and at the store's own progress
  boundaries, independent of whether a progress token was supplied.
- Failure details are capped at 20 entries and sanitized at the MCP boundary:
  known store categories pass through, raw backend, model, download and
  database error text does not.
- No new REST endpoints, no schema change, no migration. Existing search,
  retrieval, resource and status contracts are unchanged.

The embed runtime budget became part of the public SDK contract so the MCP
layer can pass it through without reimplementing batching.

## Tests

Protocol-level tests drive real `tools/list` and `tools/call` messages over an
in-memory transport against a controlled store double — no model, no GPU, no
filesystem scan. Covered: schema and annotations, option validation and
conversion, structured results, safe errors, progress with and without a token,
cross-session busy behaviour, lock release after success/error/cancellation,
bounded failure details, and transport parity for stdio and HTTP.

`test/mcp-maintenance.test.ts` (new) plus additions to `test/sdk.test.ts` and
`test/mcp.test.ts`. Node 175 passed, Bun 175 passed, `npm run test:types`,
`npm run build` and `npm run test:package` green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,25 @@
 
 ## [Unreleased]
 
+### Added
+
+- The MCP server now exposes `update` and `embed` through the same central tool
+  registration for stdio and Streamable HTTP. The documented maintenance flow
+  is `update`, inspect `status.needsEmbedding`, then `embed` only when work is
+  pending. Both tools report progress when the client supplies a progress
+  token, return structured counters, and write only to QMD's derived index.
+  Progress delivery is best-effort and cannot change a successful tool result.
+  `update` never modifies source files or executes configured update commands;
+  `embed` uses only the centrally configured model and defaults to a 30-minute
+  runtime limit. Values above the documented safe Node timer maximum are
+  rejected; `0` remains unlimited. A per-process, per-store fail-fast lock
+  permits one maintenance writer while reads remain available, and
+  cancellation/error paths release the lock. Separate processes continue to
+  rely on SQLite/WAL/busy-timeout. Busy, cancelled, failed, and partial
+  embedding results are surfaced with `isError` without adding REST maintenance
+  endpoints. MCP failure details preserve known safe Store categories and
+  sanitize raw backend, model, download, and database error text.
+
 ## [2.6.3] - 2026-06-24
 
 ### Added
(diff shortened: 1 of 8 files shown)


michabbb:
Closing in favour of a single, self-contained PR.

This branch turned out to be the base for a second piece of work — five collection management tools for MCP (`collection_list`, `collection_show`, and opt-in `collection_add`/`collection_rename`/`collection_remove`) — which builds directly on the server factory introduced here and cannot be reviewed or merged independently of it.

Rather than leaving a stacked PR that depends on this one, I will open a replacement PR containing both parts as separate commits, so the collection half can be dropped without losing the update/embed half.

Nothing here is withdrawn; it will be part of the new PR unchanged.

PR #804, opened 2026-07-29
feat(mcp): add collection management tools

michabbb:
## Problem

MCP clients can search and maintain an index, but they cannot see or change which collections exist. `status` hints at them, yet there is no way to list one, inspect its configuration, or register a directory without dropping to a shell — impossible in environments that have no shell access.

## What this adds

**This PR contains two independent parts.** The first is #802, which I closed in favour of a single self-contained PR: the second part builds directly on the server factory it introduces and cannot be reviewed or merged without it. Both are separate commits, so the second can be dropped without losing the first.

**Part 1 — `update` and `embed`** (commit `a6bc12d`, unchanged from #802): maintenance tools registered in the shared factory, with progress notifications, a fail-fast lock, cancellation and sanitized failure details.

**Part 2 — five collection tools** (commit `e12ce1f`): `collection_list` and `collection_show` are always available; `collection_add`, `collection_rename` and `collection_remove` are off by default and exist only for a trusted local stdio client that opts in with `qmd mcp --enable-collection-management`.

- `collection_list` and `collection_show` report name, path, pattern, active and indexed document counts, last change and the unscoped-search default. They list every configured collection, including ones that have never been indexed.
- `collection_add` registers a directory without indexing it, resolves the path to its real location, and rejects a missing path, a non-directory, a taken name or a duplicate path/pattern pair before the store is called.
- `collection_rename` renames the collection and its indexed `qmd://` paths.
- `collection_remove` deletes the collection and its indexed data; source files are never touched.

## Safety boundaries

- The factory takes the transport and the operator opt-in. Writing tools register only for `stdio` **and** an explicit opt-in, so the rule lives in one place and applies to any future entry point rather than depending on each caller. Over Streamable HTTP they never appear in `tools/list` — that transport has no authentication and would otherwise turn a search service into a remote filesystem administration API.
- `--enable-collection-management` is rejected together with `--http` instead of being silently ignored. There is deliberately no environment variable or config file equivalent, so the opt-in cannot be inherited by a child process.
- All writes share the existing maintenance lock with `update` and `embed`. It is taken **before** validation, not just around the mutation, so a concurrent caller cannot pass a uniqueness check that a competing write is about to invalidate. Read tools stay available.
- Store exceptions are masked. Where an operation may have partially succeeded, the message asks the caller to verify rather than suggesting a retry.

## Two SDK defects fixed along the way

The tools would otherwise expose them:

- `removeCollection` removed only the configuration entry and left the indexed documents searchable. It now removes them transactionally and reports `{ removed, deletedDocs, cleanedHashes }`.
- `renameCollection` renamed only the configuration entry, so indexed documents stayed under the old collection name. It now renames both in one transaction.

Both counters are measured with `COUNT(*)` rather than read from `.changes`, which under Bun includes changes made by the documents FTS triggers.

Separately, the orphaned-content cleanup used two contradictory rules: `removeCollection` treated only active documents as references, while `cleanupOrphanedContent` documents why inactive tombstones must count. Because `documents.hash` is a foreign key on `content.hash` with `ON DELETE CASCADE`, the active-only variant silently hard-deleted tombstones belonging to *other* collections. Both call sites now share one predicate.

## Tests

Protocol-level tests drive real `tools/list` and `tools/call` messages over an in-memory transport against a controlled store double — no model, no GPU, no filesystem scan. The transport boundary is asserted as an invariant rather than a list of tool names: enabling the opt-in must not change what HTTP exposes, which stays true for any tool added later.

Node and Bun green; `npm run test:types`, `npm run build` and `npm run test:package` green.

> Note: `test/store-paths.test.ts` fails 4 tests on Linux and macOS on `main` as well — those tests expect Git-Bash-to-Windows path conversion unconditionally. Unrelated to this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,16 @@
 
 ### Added
 
+- MCP now exposes read-only `collection_list` and `collection_show` tools on
+  both transports. Trusted local stdio clients can explicitly enable
+  `collection_add`, `collection_rename`, and destructive `collection_remove`
+  with `qmd mcp --enable-collection-management`. The writing tools are
+  deliberately excluded from HTTP, validate names and local paths, return
+  structured results, preserve source files, and share the fail-fast Store
+  maintenance lock with `update` and `embed`. The documented add workflow is
+  `collection_add`, `update`, inspect `needsEmbedding`, then `embed` only when
+  needed. Ignore patterns can be supplied during add but are not returned by
+  collection read tools.
 - The MCP server now exposes `update` and `embed` through the same central tool
   registration for stdio and Streamable HTTP. The documented maintenance flow
   is `update`, inspect `status.needsEmbedding`, then `embed` only when work is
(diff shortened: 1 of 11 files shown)


michabbb:
Closing: the transport restriction on the writing collection tools turned out to be the wrong trade-off for the intended use case. Reworking so all five tools are available over both transports, with DNS-rebinding protection on the HTTP transport instead of a blanket exclusion.

PR #815, opened 2026-08-05
perf(mcp): cache server instructions instead of rebuilding per request

fxstein:
## What

Caches the built MCP server instructions per store — shared in-flight promise, 60s TTL — instead of rebuilding them inside every `createMcpServer()` call.

## Why

The Streamable HTTP transport creates one `McpServer` per MCP session, so every `initialize` re-runs the full index-status aggregation behind `buildInstructions()`. On a large index that recomputation dominates the handshake, and agent reconnect bursts stack N status scans (observed on a production 83.5k-row index as multi-second to tens-of-seconds initializes; companion fix #814 removes most of the per-scan cost, this PR removes the per-session recomputation).

With the cache, a reconnect burst costs one status scan: concurrent initializes share a single in-flight build.

## Details

- Failed builds are dropped from the cache, so the next `initialize` retries instead of replaying a cached rejection for the rest of the TTL window.
- Index updates from other processes surface within 60s without restarting the server; the `status` tool is untouched and always live.
- Stdio transport behavior is unchanged (one server per process).

## Testing

- New test in the HTTP describe of `test/mcp.test.ts`: two fresh sessions receive identical instructions while the index is mutated between them.
- Heads-up: on a clean checkout of current `main` (e428df7) the HTTP describe fails in `beforeAll` before any test runs — `rebuildFTSForCjkNormalization` throws `SqliteError: no such column: T.name` against the raw-SQL fixture DB (pre-existing; CI skips the block via `describe.skipIf(CI)`). I verified the behavior standalone against a real `createStore` schema: a second session receives the identical cached string, still carrying the pre-mutation document count.
- `tsc -p tsconfig.build.json --noEmit` clean.

### DIFF
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -164,6 +164,34 @@ async function buildInstructions(store: QMDStore): Promise<string> {
 }
 
 /**
+ * Cache built instructions per store. The HTTP transport builds a fresh
+ * McpServer per request (sessionless MCP 2026-07-28), and buildInstructions()
+ * runs a full index-status aggregation, so rebuilding it per request puts that
+ * scan in front of every call. Concurrent builds share one in-flight promise,
+ * and entries expire after a short TTL so index updates from other processes
+ * still surface without a server restart.
+ */
+const INSTRUCTIONS_CACHE_TTL_MS = 60_000;
+const instructionsCache = new WeakMap<QMDStore, { promise: Promise<string>; builtAt: number }>();
+
+function getInstructions(store: QMDStore): Promise<string> {
+  const cached = instructionsCache.get(store);
+  if (cached && Date.now() - cached.builtAt < INSTRUCTIONS_CACHE_TTL_MS) {
+    return cached.promise;
+  }
+  const promise = buildInstructions(store);
+  instructionsCache.set(store, { promise, builtAt: Date.now() });
+  // Drop failed builds so the next initialize retries instead of replaying
+  // the cached rejection for the rest of the TTL window.
+  promise.catch(() => {
+    if (instructionsCache.get(store)?.promise === promise) {
+      instructionsCache.delete(store);
+    }
+  });
+  return promise;
+}
+
+/**
  * Create an MCP server with all QMD tools, resources, and prompts registered.
  * Shared by both stdio and HTTP transports.
  */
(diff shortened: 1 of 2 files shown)


fxstein:
Rebased onto current `main` (`dbfd0b4`). The branch was unmergeable against the sessionless MCP rework; it now applies cleanly.

The rework strengthens the premise rather than weakening it. `createMcpHandler(() => createMcpServer(store), ...)` builds a fresh `McpServer` for **every HTTP request**, not once per session, so `getInstructions(store)` is re-derived on each call instead of once per client connection. The original title said "per session" and I have updated it to "per request" to match.

The rebase keeps upstream's `cacheHints` block verbatim and only replaces the `instructions:` value with the cached one:

```ts
const server = new McpServer(
  { name: "qmd", version: getPackageVersion() },
  {
    instructions: await getInstructions(store),
    cacheHints: {
      "tools/list": { ttlMs: 60_000, cacheScope: "private" },
      "server/discover": { ttlMs: 60_000, cacheScope: "private" },
      "resources/read": { ttlMs: 0, cacheScope: "private" },
    },
  },
);
```

One test in `test/mcp.test.ts` had to be replaced: it drove the old `Mcp-Session-Id` handshake, which no longer exists. The replacement is sessionless and tolerates either a JSON or an SSE-framed response body.

Heads-up on the suite, unrelated to this branch: `POST /mcp initialize still works for 2025-era clients` fails on clean `main` as well, expecting `application/json` and getting `text/event-stream`. I confirmed that by running the suite on `main` with no patch applied before touching anything here. On this branch the result is the same single pre-existing failure, 66 passed.

fxstein:
Correcting my note above, because CI going green on this branch reads as contradicting it and I want the record straight.

The failure is real and it is on `main`, but **CI never runs it**. `describe("MCP HTTP Transport")` at `test/mcp.test.ts:937` is `skipIf(!!process.env.CI)`, and the workflow sets `CI: true`, so the whole block is skipped in every Bun job. Measured on a clean checkout of `dbfd0b4` with a real `bun install --frozen-lockfile`, using the exact CI command:

```
bun test --timeout 60000 --preload ./src/test-preload.ts test/mcp.test.ts
  65 pass, 1 fail
  POST /mcp initialize still works for 2025-era clients (no session id)
  expect(received).toContain("application/json")
  Received: "text/event-stream"

CI=true bun test --timeout 60000 --preload ./src/test-preload.ts test/mcp.test.ts
  52 pass, 16 skip, 0 fail
```

Same commit, same command, the only difference is `CI`. So the green checks on this PR and on `main` are both consistent with the test being broken.

The consequence for this PR, stated plainly: my replacement for the obsolete session-handshake test lives in that same `skipIf(CI)` block, so it does not run in CI either. It passes locally under the first invocation above. If you would rather have it somewhere CI actually exercises, say where and I will move it.

The `2025-era clients` assertion itself looks like it outlived the sessionless rework rather than being a live defect: it wants `application/json` from an `initialize` that the server now answers as `text/event-stream`. That is your call to make, not something I have changed here.

PR #818, opened 2026-08-05
fix(store): keep caller intent out of query expansion and drop dud cached expansions

fxstein:
## What

Two coupled fixes to auto-expansion:

1. `expandQuery()` consumes **only the query text** — caller `intent` no longer enters the expansion prompt or the expansion cache key. Intent keeps doing its real jobs downstream: the reranker query prefix and keyword-based chunk/snippet selection (both unchanged).
2. `hybridQuery()` deletes a cached expansion whose runnable sub-queries all came back empty, so a dud expansion can't deterministically replay the same misses on every warm repeat of the query.

## Why

When `intent` is passed to the MCP `query` tool — which the server's own instructions tell every client to always do — the expansion model builds sub-queries from the *intent's* meta-language instead of the query. Production cache rows (agent MCP clients, 83.5k-doc index) captured it directly; the worst case was an expansion whose **every** lex/vec/hyde line was the caller's intent string verbatim:

```
[{"type":"lex","query":"post-purge retest of the exact poisoned auto-expand query"},
 {"type":"vec","query":"post-purge retest of the exact poisoned auto-expand query"}, ...]
```

Degenerate terms match nothing → empty results; the dud is cached per `(query, intent)` → deterministic empties on every warm repeat. Control on the same index: the identical query returned **20 relevant hits without intent, zero with it**. The `hasQueryTerm` output filter doesn't catch this because real-world intents share vocabulary with their queries.

The finetune data does condition on intent, but only in a small slice of examples with short task-style intents ("optimizing a C++ application"); free-form agent intents are far outside that distribution and get reproduced verbatim.

After this fix, the exact production repro (same query, same intent, same index) returns **5 relevant hits in 8.9s** where it previously returned zero — and dropping intent from the cache key orphans all previously poisoned entries without needing a purge.

## Details

- The empty-contribution invalidation only judges sub-queries the store could actually run: on FTS-only stores, vec/hyde expansions never execute and say nothing about the expansion's quality (guards against invalidate/regenerate thrash).
- `ExpandQueryOptions.intent` is deprecated (accepted, ignored) for API compatibility; `SearchOptions.intent` is unchanged. The instructions' "always provide `intent`" guidance becomes safe again with this in place.
- If you'd rather keep intent-conditioned expansion and instead harden the output filter or the finetune distribution, happy to rework — this shape is the smallest change that stops the poisoning.

## Testing

- New tests in `test/store.test.ts`: cached expansions serve without invoking the LLM; a cached expansion contributing nothing is dropped by `hybridQuery`; a contributing one survives. 226/226 pass; `test/structured-search.test.ts` 75/75; `tsc -p tsconfig.build.json --noEmit` clean.
- Live verification against the production index that surfaced the bug: the previously poisoned `(query, intent)` pair now returns 5 hits in 8.9s via `store.search`.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -192,6 +192,12 @@ export interface VectorSearchOptions {
  * Options for expandQuery() — manual query expansion.
  */
 export interface ExpandQueryOptions {
+  /**
+   * @deprecated Ignored. Intent no longer feeds the expansion model — caller
+   * intent is meta-language the model reproduced verbatim as sub-queries.
+   * Pass intent via SearchOptions instead, where it shapes reranking and
+   * snippet selection.
+   */
   intent?: string;
 }
 
@@ -427,7 +433,7 @@ export async function createStore(options: StoreOptions): Promise<QMDStore> {
     },
     searchLex: async (q, opts) => internal.searchFTS(q, opts?.limit, opts?.collection),
     searchVector: async (q, opts) => internal.searchVec(q, llm.embedModelName, opts?.limit, opts?.collection),
-    expandQuery: async (q, opts) => internal.expandQuery(q, undefined, opts?.intent),
+    expandQuery: async (q) => internal.expandQuery(q),
     get: async (pathOrDocid, opts) => internal.findDocument(pathOrDocid, opts),
     getDocumentBody: async (pathOrDocid, opts) => {
       const result = internal.findDocument(pathOrDocid, { includeBody: false });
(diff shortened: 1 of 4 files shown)


tobi:
Rebased onto current main as #870 (original branch still conflicted). Same two fixes: intent stays out of expansion/cache, dud cached expansions are dropped. Will close this once #870 merges.

tobi:
Landed on main via #870 (rebased onto current main; this branch was still conflicted).

fxstein:
Thank you!

Issue #735, opened 2026-06-14
qmd embed deadlocks at 0% CPU on Apple M5 Max + macOS 26.4 — Metal shader JIT compile failure in bundled llama.cpp b8390 (distinct from #673/#724)


e3322-pcsk9:
## Summary

On **Apple M5 Max + macOS 26.4 Tahoe**, `qmd embed` (and any path that initializes the Metal backend for batch embedding) **deadlocks at 0% CPU** inside `libggml-metal` after `ggml_metal_library_init_from_source: error compiling source`. This is **not** the same root cause as #673 / #724 (those cover the `maxDuration` timeout on *slow-but-progressing* runs). Here the process never makes any progress at all — it hangs in `kevent` waiting on a native async call that never resolves. Patching `maxDuration` does **not** help (verified).

## Environment

| Component | Version |
|---|---|
| qmd | 2.5.3 (`@tobilu/qmd` via bun global) |
| macOS | 26.4 (Build 25E246) |
| Chip | **Apple M5 Max** (Mac17,6; 107.5 GB unified memory) |
| node-llama-cpp | 3.18.1 (bundled) |
| llama.cpp | b8390 (`_nlcBuildMetadata.json`: platform macOS 25.3.0 / arm64 / metal) |
| Embed model | `Qwen3-Embedding-8B-f16.gguf` (14 GB) — also reproduces with the smaller default after `models.embed` is removed |
| Xcode CLT | clang 21.0.0, SDK 26.5 |
| `GGML_METAL_NO_RESIDENCY` | `1` (set by `bin/qmd` launcher per #22593 mitigation) |

## Reproduction

```bash
qmd embed -f              # or: qmd embed --max-docs-per-batch 1 --max-batch-mb 1
```

### Observed output (stays here forever)

```
Model: /Users/.../Qwen3-Embedding-8B-f16.gguf
Batch: N docs / X MB
[node-llama-cpp] ggml_metal_library_init_from_source: error compiling source
█░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░░  ← progress bar frozen at 0%
```

### Process state after 5 minutes

```
$ ps -p <PID> -o pid,stat,etime,pcpu,pmem
  PID STAT ELAPSED  %CPU %MEM
 NNNN SN    05:01    0.0 13.2     ← 13.2% MEM = 17.7 GB model loaded, but 0% CPU
```

### `sample <PID> 3` call graph (main thread)

```
[...]

PR #821, opened 2026-08-10
fix(embed): session reliability — 120min cap, configured-model pending count, node-llama-cpp 3.19.1

e3322-pcsk9:
## Summary

Three embed-pipeline reliability fixes, all surfaced by one real-world failure on macOS 26.4 (Apple M5 Max) where `qmd embed` died at exactly 30m08s with `⚠ Session expired — skipping 834 remaining chunks`:

1. **`fix(update)`** — post-update "N unique hashes need vectors" notice counted pending hashes against the *built-in default* embed model instead of the configured one, so anyone overriding `models.embed` / `QMD_EMBED_MODEL` got a false full-corpus warning after every `qmd update`.
2. **`fix(embed)`** — the default 30-minute embed session cap silently truncated full re-embeds of mid-size indexes (~1.5k docs with a 4B model needs ~40 min); remaining chunks were recorded as "LLM session expired before embedding chunk".
3. **`chore(deps)`** — node-llama-cpp 3.18.1 → 3.19.1 (llama.cpp b8390 → b10068), fixing a sporadic embed deadlock on Apple Silicon + macOS 26.4.

## Root causes & evidence

**Session expiry (cap).** `DEFAULT_EMBED_MAX_DURATION_MS` was 30 min; the observed run aborted at 30m08s exactly. The session wrapper (`withLLMSessionForLlm`) has no auto-renewal, so the remaining 834 chunks were dropped. `--timeout <minutes>` already existed to override it, but the default was too small for the corpora people actually have.

**False pending count.** After the backlog was cleared (`qmd doctor`: all 1,441 docs on the current fingerprint, 0 pending), `qmd update` still printed *"1263 unique hashes need vectors"* — and `COUNT(DISTINCT hash)` of active documents was exactly 1263. `updateCollections()`/`indexFiles()` called `getHashesNeedingEmbedding(db)` with no model argument; `showStatus()` and `embed` already resolve the configured model correctly. Both call sites fixed.

**Metal deadlock.** On macOS 26.4, llama.cpp's Metal 4 tensor-API self-test probe fails to JIT-compile (`ggml_metal_library_init_from_source: error compiling source`; matmul2d_descriptor static_assert in the 26.4 SDK's MetalPerformancePrimitives headers — see llama.cpp#21048 and #735). On b8390 (node-llama-cpp 3.18.1) this failure is ungraceful: we reproduced the reported hang (node-llama-cpp#595) — 15 wall minutes at 0% CPU after model load. On b10068 (3.19.1) the same probe failure degrades gracefully: Metal stays enabled, tensor API disables cleanly, no deadlock. Probe still fails upstream, so M5 tensor acceleration is not yet enabled — a future llama.cpp bump should recover a 2-3× prefill speedup.

## Verification

- `npm test` fully green: vitest 881 passed / 0 failed, bun 881 passed / 0 failed, package smoke (build + grammars + compiled CLI under Node and Bun) OK.
- `qmd doctor` post-upgrade: Metal backend enabled, **3 sampled chunks reproduce stored vectors** (fingerprint-compatible, no re-embed needed).
- Real `qmd update`: false pending notice gone.
- Real embed backlog completed (81 docs; fingerprint count 1,360 → 1,441), plus end-to-end `qmd query` smoke against newly embedded docs.

## Notes

- The bun `trustedDependencies` entry records the postinstall trust so `bun install` fetches the prebuilt binary without manual `bun pm trust`.
- The Metal probe error line still prints on stderr on macOS 26.4; it is cosmetic (graceful degradation), tracked upstream.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,34 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Default embed session cap (`DEFAULT_EMBED_MAX_DURATION_MS`) raised from 30 to
+  120 minutes. A full re-embed of a mid-size index (e.g. ~1,500 documents with a
+  4B embedding model) takes ~40 minutes and was silently truncated at the
+  30-minute mark, leaving remaining chunks marked "LLM session expired". The
+  `--timeout <minutes>` flag still overrides the cap and `--timeout 0` removes
+  it; re-running `qmd embed` resumes where a capped run left off.
+- `node-llama-cpp` upgraded 3.18.1 → 3.19.1 (llama.cpp b8390 → b10068). On
+  Apple Silicon running macOS 26.4, the Metal 4 tensor-API self-test probe
+  fails to JIT-compile (upstream llama.cpp#21048 fixed the probe dimensions,
+  but the probe still fails on M5 + macOS 26.4 — see tobi/qmd#735). On b8390
+  this failure was ungraceful: embed could deadlock at 0% CPU mid-run
+  (node-llama-cpp#595). On b10068 the same probe failure degrades gracefully —
+  Metal stays enabled, tensor API disables cleanly, no deadlock. Stored vectors
+  verified compatible post-upgrade (`qmd doctor` vector-sample check passes).
+- README/CLAUDE.md: documented `qmd embed --timeout` alongside the memory flags
+  and updated CLI help text for the new default.
+
+### Fixed
+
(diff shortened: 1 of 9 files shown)


tobi:
The `node-llama-cpp` bump in this PR is superseded by #891 (3.20.0). Embed-session reliability still needs a rebase onto main without the llama dep bump.

PR #836, opened 2026-08-12
fix(mcp): shut down gracefully on stdin EOF instead of orphaning to PID 1 (#751)

### LINKED ISSUE #751: qmd mcp (stdio) doesn't exit on stdin EOF — orphans to PID 1 and leaks RAM after parent dies

tobi:
Relands #767 onto current main (original was conflicted).

Fixes #751.

Original author: @fa1k3

## Summary
Shut down the MCP stdio server gracefully on stdin EOF instead of orphaning to PID 1.

Changelog conflict resolved by keeping all Unreleased bullets from both sides.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -97,6 +97,15 @@
   header, always printed with `docid`. Column positions are now stable across
   runs and formats.
 
+### Fixed
+
+- `qmd mcp` (stdio) now shuts down gracefully when stdin reaches EOF instead
+  of orphaning to PID 1 when the parent MCP client dies (#751): the server
+  closes its transport, gives in-flight request handlers a bounded window to
+  settle, closes the store (which disposes its llama.cpp instance), and lets
+  the process drain via `process.exitCode` (no forced `process.exit()`, which
+  has caused exit-time native crashes before).
+
 ## [2.6.3] - 2026-06-24
 
 ### Added
--- a/src/mcp/server.ts
+++ b/src/mcp/server.ts
@@ -168,7 +168,10 @@ async function buildInstructions(store: QMDStore): Promise<string> {
  * Create an MCP server with all QMD tools, resources, and prompts registered.
  * Shared by both stdio and HTTP transports.
  */
-async function createMcpServer(store: QMDStore): Promise<McpServer> {
+async function createMcpServer(store: QMDStore, inflight?: InflightGate): Promise<McpServer> {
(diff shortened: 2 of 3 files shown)


tobi:
Superseded: #767 and #837 already landed the same fix on main.

PR #779, opened 2026-07-20
fix: make prepare script cross-platform so npm install works on Windows

sv789:
Fixes #778.

`prepare` was `[ -d .git ] && ./scripts/install-hooks.sh || true`, which cmd.exe can't run — `npm install` from a git checkout fails on Windows before completing (`'[' is not recognized as an internal or external command`).

This ports `scripts/install-hooks.sh` to a cross-platform `scripts/install-hooks.mjs` (same behavior: copy `scripts/pre-push` into `.git/hooks`, skip when not in a git checkout) and points `prepare` at it. The now-unused `.sh` is removed.

Verified on Windows 11 / Node v24.16.0: `npm install` completes, `npm run prepare` prints `Installed git hooks: pre-push`, and the hook lands in `.git/hooks/`. macOS/Linux behavior is unchanged (same copy + chmod).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #841, opened 2026-08-13
fix: make prepare script cross-platform for Windows (#778)

### LINKED ISSUE #778: npm install fails on Windows: prepare script is POSIX-only
**Environment:** Windows 11 Pro (10.0.26200), Node v24.16.0 / npm 11, fresh `git clone`, v2.6.3 (`e428df7`).

`npm install` fails before completing:

```
> @tobilu/qmd@2.6.3 prepare
> [ -d .git ] && ./scripts/install-hooks.sh || true

'[' is not recognized as an internal or external command,
operable program or batch file.
'true' is not recognized as an internal or external command,
operable program or batch file.
npm error code 1
```

npm runs lifecycle scripts through cmd.exe on Windows, which has no `[`, no `./x.sh` execution, and no `true` — so every Windows contributor installing from a git checkout hits this. Published-package consumers are unaffected (`prepare` only runs in checkouts), and bun users are unaffected too, which is probably why it hasn't been reported.

**Workaround:** `npm install --script-shell "C:\Program Files\Git\bin\bash.exe"`

**Suggested fix:** port the 3-line hook installer to a cross-platform Node script and call that from `prepare` — PR incoming.

tobi:
## Summary

Relands #779 onto current `main`, preserving #824 build step in `prepare`.

Fixes #778. Hook install is now `scripts/install-hooks.mjs` (cross-platform Node). Prepare is:

```
node scripts/install-hooks.mjs && node scripts/build.mjs
```

## Test plan

- [x] vitest `test/install-hooks.test.ts` — 3/3 pass
- [x] bun test `test/install-hooks.test.ts` — 3/3 pass
- [x] `node scripts/install-hooks.mjs` installs pre-push
- [x] package-smoke asserts prepare is Windows-safe and still builds dist

Supersedes #779.


tobi:
Closing: #779 already relanded via #839 + #840 on main (this branch conflicted as a duplicate).

PR #842, opened 2026-08-13
fix: make prepare Windows-safe while still building dist

### LINKED ISSUE #778: npm install fails on Windows: prepare script is POSIX-only

tobi:
Fixes #778.

Relands #779 by @sv789 onto current main. Makes prepare Windows-safe via Node install-hooks.mjs while keeping the dist build step from #824.

Local tests green: package.test.ts, package-smoke, full bun test suite.


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

tobi:
Closing as duplicate: #779 was already relanded via #839 (plus #840 tests/changelog) while this branch was in flight. Current main already has prepare = node scripts/install-hooks.mjs && node scripts/build.mjs.

Issue #743, opened 2026-06-22
Feature Request: Allow host configuration for qmd mcp --http (Docker support)


alessandroperrone-coder:
Hi team,

First of all, thank you so much for the great work you are doing on this project!

I am reaching out because I am trying to containerize your service inside a Docker container using the following command:

```Bash
qmd mcp --http
```

Currently, the server binds exclusively to the loopback interface (127.0.0.1), which prevents accessing the service from outside the container. Looking at the source code (in server.ts at line 842), I noticed that the host is hardcoded to "localhost":

```TypeScript
// Current code (server.ts, line 842)
await new Promise<void>((resolve, reject) => {
    httpServer.on("error", reject);
    httpServer.listen(port, "localhost", () => resolve());
});
```
Proposed Change
It would be extremely helpful to be able to configure the host address (for instance, setting it to 0.0.0.0 for Docker), just like we currently can for the port. The code modification could be straightforward by introducing a host variable:

```TypeScript
// Proposed modification
await new Promise<void>((resolve, reject) => {
    httpServer.on("error", reject);
    httpServer.listen(port, host, () => resolve());
});
```
It would be amazing to be able to pass this parameter via a CLI flag (e.g., --host 0.0.0.0) or an environment variable.

Thank you again for your support and keep up the great work!

unfall103-debug:
我也不能理解很多开发者硬编码host...都是单机用户？

PR #885, opened 2026-08-13
fix(mcp): validate Origin and Host on the HTTP transport (DNS rebinding)

### LINKED ISSUE #881: MCP HTTP server accepts any Origin/Host — DNS rebinding gives any website full read access to the indexed corpus
`qmd mcp --http` performs no `Origin` or `Host` validation, so any website the user visits can read the entire indexed corpus via DNS rebinding.

Reproduced on `main` (v2.6.3).

## Problem

`startMcpHttpServer()` in [`src/mcp/server.ts`](https://github.com/tobi/qmd/blob/main/src/mcp/server.ts) serves three unauthenticated routes — `POST /mcp`, `POST /query`, `POST /search` (plus `GET /health`) — and never inspects the `Origin` or `Host` request headers. The MCP spec's Streamable HTTP section requires local servers to validate `Origin` precisely to prevent DNS rebinding, and the SDK transport ships the knob for it (`enableDnsRebindingProtection` / `allowedHosts` / `allowedOrigins` in `WebStandardStreamableHTTPServerTransport`) — QMD constructs the transport without any of them:

```ts
const transport = new WebStandardStreamableHTTPServerTransport({
  sessionIdGenerator: () => randomUUID(),
  enableJsonResponse: true,
  onsessioninitialized: (sessionId: string) => { ... },
});
```

The REST endpoints (`/query`, `/search`) bypass the transport entirely, so even enabling the SDK flag would not cover them — the check has to live in the `createServer()` handler.

### Why the loopback bind is not a mitigation

`localhost` binding stops *remote* attackers, not *browser* ones. A page the user visits serves a hostname whose DNS record has a short TTL and initially points at the attacker's server; after the page loads it re-resolves to `127.0.0.1`. From then on the browser considers `http://attacker.example:8181` same-origin with the attacker's page, and `fetch()` results are fully readable by attacker JavaScript. The only thing that would have stopped it is the server rejecting the request because `Origin`/`Host` do not name a loopback address — which QMD never checks.

The `--host` / `QMD_HOST` option (#743) makes this materially worse: a container or LAN bind on `0.0.0.0` exposes the same unauthenticated corpus-read API to everyone on the network.

## Impact

Full read access to every document QMD has indexed — typically the user's notes, journals, and private repos — plus the collection layout and on-disk paths leaked through `status`/`instructions`. It is a silent exfiltration primitive: no prompt, no log the user would notice, and it works for as long as the daemon (`qmd mcp --http --daemon`) is running.

## Repro

Server:

```bash
qmd mcp --http --port 8181
```

Attacker-controlled headers, from any other process (the same request a rebound browser page makes):

```bash
curl -s -X POST http://localhost:8181/query \
  -H 'Content-Type: application/json' \
  -H 'Origin: https://evil.example' \
  -H 'Host: evil.example' \
  -d '{"searches":[{"type":"lex","query":"password"}]}'
```

Observed (against a scratch index seeded with one document):

```
status: 200
{"results":[{"docid":"#h1","file":"qmd://docs/secrets.md","title":"Secrets","score":1,
  "snippet":"5: My AWS root password is hunter2 and my recovery codes are 111-222.\n"}]}
```

`POST /mcp` behaves the same — `initialize` succeeds with `Origin: https://evil.example`, returning the server instructions (document count, collection names) and a session id usable for `tools/call`.

## Proposed fix

Validate both headers in the `createServer()` handler, before any routing, so `/mcp` and the REST endpoints are covered by one gate:

1. **Origin.** Absent → allow (curl, MCP stdio-style clients, and Claude Code's HTTP client send no `Origin`; browsers always do). Present → must parse to a loopback host (`localhost`, `127.0.0.0/8`, `[::1]`) or appear in an explicit allowlist. Otherwise `403` with a JSON-RPC-shaped error body.
2. **Host.** When the server is bound to a loopback address (the default), the `Host` header's hostname must also be loopback or allowlisted. When the operator has deliberately bound elsewhere (`--host 0.0.0.0` for Docker), skip the host check — the legitimate `Host` value is then unknowable — unless an allowlist is configured.
3. **Escape hatches.** `QMD_ALLOWED_ORIGINS` / `QMD_ALLOWED_HOSTS` (comma-separated) to extend the allowlist, and `QMD_ALLOWED_ORIGINS=*` to opt out entirely for users who front QMD with their own auth proxy.

Defaults stay backwards compatible for every non-browser client, since those send no `Origin` and a loopback `Host`.

Additionally, log a one-line warning at startup when binding to a non-loopback host without an allowlist, so the Docker path is an informed choice rather than a silent one.

I have this implemented with unit tests for the header predicates and integration tests that drive a live server with spoofed headers; PR to follow.

Chirag6722:
Closes #881.

## What was wrong

`startMcpHttpServer` served `POST /mcp`, `POST /query` and `POST /search` without ever looking at `Origin` or `Host`. Binding to `localhost` keeps *remote* clients out; it does nothing about the user's own browser. A page served from `attacker.example` can re-point that hostname at `127.0.0.1` after load — at which point the browser considers `http://attacker.example:8181` same-origin with the page and hands the full response body to attacker JavaScript. Every indexed document, silently, for as long as `qmd mcp --http --daemon` is running.

Verified against `main` before the fix — `Origin: https://evil.example` returned `200` with document snippets from `/query`, and `initialize` succeeded on `/mcp`.

## The fix

The check lives in the `createServer` handler, ahead of routing. That placement matters: the REST endpoints bypass the MCP transport entirely, so the SDK's own `enableDnsRebindingProtection` flag would have covered `/mcp` and left `/query` wide open.

- **`Origin` absent → allowed.** curl, MCP clients and editors omit the header; browsers cannot. Absence is not the signal — a foreign value is. This is what keeps every existing local client working unchanged.
- **`Origin` present → must be loopback** (`localhost`, `*.localhost`, `127.0.0.0/8`, `::1`, `::ffff:127.0.0.1`) or explicitly allowlisted. Otherwise `403`.
- **`Host` enforced when the bind address names a concrete interface.** For a wildcard bind (`--host 0.0.0.0`, the Docker case from #743) the legitimate `Host` is unknowable, so the check is skipped and startup warns; setting `QMD_ALLOWED_HOSTS` re-enables it. A concrete non-loopback bind (`--host 192.168.1.5`) trusts its own address automatically.
- **`0.0.0.0` is not treated as loopback.** Browsers route it to the loopback interface on Linux and macOS, which makes it a rebinding-free path to a local server, so it stays untrusted as an origin.
- **Escape hatches:** `QMD_ALLOWED_ORIGINS` / `QMD_ALLOWED_HOSTS` (comma-separated) extend the allowlists; `QMD_ALLOWED_ORIGINS=*` disables the screen for anyone fronting QMD with their own auth proxy, and warns at startup.

The predicates live in `src/mcp/origin-guard.ts` as pure functions so they are testable without a socket.

## Tests

New `test/mcp-origin-guard.test.ts` — 35 tests, all passing:

- Unit coverage of the hostname/origin/host predicates, including the near-misses that a naive substring check would let through (`localhost.evil.example`, `127.0.0.1.evil.example`, `https://evil.example#localhost`), the opaque `null` origin from sandboxed frames, non-http schemes, and `0.0.0.0`.
- Guard resolution across bind modes: loopback, wildcard, wildcard + allowlist, concrete LAN address, `*`.
- Live-server integration: boots `startMcpHttpServer` on an ephemeral port against a seeded scratch index and asserts `403` (and no leaked document text) for a foreign `Origin` on `/query`, `/search` and `/mcp`; `403` for a rebound `Host` sent over a raw socket, since `fetch()` forbids setting that header; and `200` for the paths that must keep working — no `Origin`, a loopback `Origin`, and `/health`.

Existing MCP HTTP transport tests still pass (`initialize`, `tools/list`, `tools/call get`, absolute line numbers, session TTL, REST URI prefix, `/health`, 404). `tools/call query` is the one test I could not run here — it needs the query-expansion GGUF and the HF CDN 403s mid-download on this machine; it is unrelated to this change.

`tsc -p tsconfig.build.json --noEmit` is clean.

## Drive-by

`test/mcp.test.ts` had two hardcoded `/tmp/...` database paths, which made the whole file fail at `beforeAll` on Windows with `SQLITE_CANTOPEN` — including the HTTP transport block I needed as a regression baseline. Both now use `tmpdir()`. No behaviour change on POSIX.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,20 @@
 
 ## [Unreleased]
 
+### Security
+
+- `qmd mcp --http` now validates the `Origin` and `Host` headers on every
+  request and answers `403` when they name anything but a loopback address
+  (#881). Without this, any web page the user visited could re-point its own
+  hostname at `127.0.0.1` (DNS rebinding) and read the entire indexed corpus
+  through `POST /query` or `POST /mcp` — binding to localhost is no defence
+  against a request made by the user's own browser. Requests with no `Origin`
+  header (curl, MCP clients, editors) are unaffected. Extend the allowlists with
+  `QMD_ALLOWED_ORIGINS` / `QMD_ALLOWED_HOSTS`, or set `QMD_ALLOWED_ORIGINS=*`
+  to opt out behind your own authenticating proxy. A wildcard bind
+  (`--host 0.0.0.0`) skips the host check, since the legitimate `Host` is then
+  unknowable, and warns at startup.
+
 ### Changed
 
 - `qmd pull` (and implicit model downloads in `embed`/`query`) no longer print
--- a/README.md
+++ b/README.md
@@ -153,6 +153,28 @@ The HTTP server exposes two endpoints:
 - `POST /mcp` — MCP Streamable HTTP (JSON responses, stateless)
 - `GET /health` — liveness check with uptime
 
+##### Origin and Host validation
+
+Every request is screened before routing: a request carrying an `Origin` header
+that does not name a loopback address is rejected with `403`, as is a `Host`
+header naming something other than the address the server is bound to. This is
+what stops a web page you visit from reading your index through DNS rebinding —
+loopback binding alone does not, since the browser makes the request from your
+own machine.
+
+Requests without an `Origin` header — curl, MCP clients, editors — are
+unaffected, which covers every normal local client.
+
+| Variable | Effect |
+|----------|--------|
+| `QMD_ALLOWED_ORIGINS` | Comma-separated origins to accept in addition to loopback, e.g. `https://notes.internal`. Set to `*` to disable the check entirely. |
+| `QMD_ALLOWED_HOSTS` | Comma-separated `Host` values to accept in addition to loopback and the bind address. |
+
+`--host 0.0.0.0` cannot know which `Host` values are legitimate, so it skips the
+host check and warns at startup. Set `QMD_ALLOWED_HOSTS` to re-enable it, and
+remember the endpoints are unauthenticated — put your own auth in front of a
+server that is reachable off-host.
+
 LLM models stay loaded in VRAM across requests. Embedding/reranking contexts are disposed after 5 min idle and transparently recreated on the next request (~1s penalty, models remain loaded).
 
 Point any MCP client at `http://localhost:8181/mcp` to connect.
--- a/src/mcp/origin-guard.ts
+++ b/src/mcp/origin-guard.ts
@@ -1,0 +1,167 @@
+/**
+ * origin-guard.ts — DNS-rebinding protection for the HTTP MCP transport.
+ *
+ * Binding to localhost keeps *remote* clients out, but it does not keep a web
+ * page out. A page served from `attacker.example` can re-point that hostname at
+ * `127.0.0.1` after load (DNS rebinding); the browser then treats
+ * `http://attacker.example:8181` as same-origin with the page and hands the
+ * response body to attacker JavaScript. The only server-side signal that
+ * distinguishes such a request from a legitimate local client is the `Origin`
+ * and `Host` headers, so the MCP spec requires local HTTP servers to validate
+ * them.
+ *
+ * The predicates here are pure so they can be unit-tested without a socket;
+ * `startMcpHttpServer` applies them to every request before routing.
+ */
+
+/** Wildcard bind addresses — the request's legitimate `Host` is unknowable. */
+const WILDCARD_BIND_HOSTS = new Set(["", "0.0.0.0", "::", "[::]", "*"]);
+
+/**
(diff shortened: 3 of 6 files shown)


tobi:
Reviewed: the Origin/Host predicates in `origin-guard.ts` are sufficient. This PR cannot land as-is — it is based on the pre-#883 session transport and conflicts with current main (SDK 2.0 / `createMcpHandler`).

Relanded the same check on current main in https://github.com/tobi/qmd/pull/888 (applied in `createServer` ahead of routing so `/query` and `/search` stay covered). Closing this as superseded once 888 is green.

tobi:
Superseded by #888, which landed Origin/Host validation on the MCP HTTP transport plus index path containment. Closing as duplicate.

Issue #886, opened 2026-08-13
Cloning a repo with a .qmd/index.yml and running `qmd update` executes its `update:` hook — no trust prompt


Chirag6722:
Cloning a repository that ships a `.qmd/index.yml` and running `qmd update` inside it executes shell commands chosen by whoever wrote that file. No prompt, no trust record, no way to see it coming.

This is the `update:`-field concern from the (stale-closed) #69 audit, but with the vector that makes it actually exploitable — project-local config discovery, which landed after that audit was written. Filing fresh with a current repro as requested there.

Reproduced on `main` (v2.6.3).

## Root cause

Two features compose into an RCE that neither has on its own.

**1. `update:` is executed as a shell command.** [`src/cli/qmd.ts`](https://github.com/tobi/qmd/blob/main/src/cli/qmd.ts) in `updateCollections()`:

```ts
const yamlCol = getCollectionFromYaml(col.name);
if (yamlCol?.update) {
  console.log(`${c.dim}    Running update command: ${yamlCol.update}${c.reset}`);
  const proc = nodeSpawn("bash", ["-c", yamlCol.update], { cwd: col.pwd, ... });
```

Intended for `git pull --ff-only`, and perfectly reasonable when the user typed it via `qmd collection update-cmd`.

**2. Config is discovered by walking up from the working directory.** `findLocalConfigPath()` in [`src/collections.ts`](https://github.com/tobi/qmd/blob/main/src/collections.ts) climbs from `process.cwd()` looking for `.qmd/index.yaml` / `.qmd/index.yml`, and the CLI adopts whatever it finds:

```ts
const localConfigPath = findLocalConfigPath();
if (localConfigPath) {
  setConfigSource({ configPath: localConfigPath });
  storeDbPathOverride = getLocalDbPath(localConfigPath);
```

`getStore()` then syncs that YAML into a fresh local `index.sqlite`, so the collections — and their hooks — are live on the very first command in a freshly cloned tree. Nothing needs to have been indexed beforehand.

The result is that the *code to run* and the *decision to run it* both come from a file that arrives with the repository. `.qmd/index.yml` is a checked-in project file by design — the README documents `qmd init` as the way to create it — so it travels with the clone exactly like `.envrc` or `.vscode/tasks.json`, both of which are gated behind an explicit trust step for precisely this reason.

## Repro

```sh
mkdir -p /tmp/evil-repo/.qmd /tmp/evil-repo/docs
[...]

PR #893, opened 2026-08-13
fix(security): gate project-local collection paths and model URIs

### LINKED ISSUE #889: Project-local .qmd/index.yml can set collection paths and model URIs without a trust gate
## Context

#886 / #887 gate `update:` hooks from a checked-in `.qmd/index.yml`. The same file can still set:

- `collections.*.path` to any directory the process can read
- `models.embed` / `models.rerank` / `models.generate` to any `hf:` repo or local GGUF path

`qmd update` / `qmd embed` in a freshly cloned tree will then index that path and download/load those models, with no prompt. Global `~/.config/qmd` is unaffected.

## Ask

Reuse the #887 trust digest (or an equivalent per-config approval) for collection `path` and `models.*` when the config is project-local. Until then, treat a checked-in `.qmd/index.yml` as able to choose what gets indexed and which GGUF is loaded.

Related: #888 contains glob/symlink containment inside a collection root; it does not constrain which root a local config may name.

tobi:
Closes #889.

## What was wrong

#887 gates `update:` hooks from a checked-in `.qmd/index.yml`. The same file could still set `collections.*.path` to any directory the process can read, and `models.embed` / `models.rerank` / `models.generate` to any `hf:` repo or local GGUF. `qmd update` / `qmd embed` in a freshly cloned tree then indexed that path and downloaded/loaded those models, with no prompt. Global `~/.config/qmd` is unaffected. #888 contains glob/symlink escapes relative to the collection root; it does not constrain which root a local config may name.

## The fix

Reuse the #887 trust digest (`<config dir>/trusted.json`, `qmd trust`, `QMD_TRUST_UPDATE_HOOKS=1`).

- **Collection paths** that resolve *outside* the project (absolute dirs, `../`, `~/...`) are gated. In-project paths like `./docs` still index without approval.
- **`models.*`** that are not the built-in defaults are gated. Untrusted URIs fall back to those defaults (or `QMD_*_MODEL`); the YAML is not rewritten.
- On a TTY, `qmd update` lists the gated values and asks. Non-interactive: skip them and keep indexing in-project collections.
- A hooks-only digest is unchanged, so existing `qmd trust` records still match. Adding an out-of-project path or custom model URI changes the digest and re-arms the gate.
- `qmd collection add` / `qmd collection update-cmd` still record trust as they write, since the user just typed that path or command.

Left alone: remote embed/rerank, `qmd://`, prepare-hook, #821.

## Tests

`test/config-path-model-trust.test.ts` covers path resolution (relative to project root, `~`, symlink escape), `collectGatedSurfaces`, digest stability/sensitivity, and `decideConfigGate`.

`test/config-path-model-trust-cli.test.ts` spawns real `qmd update`/`qmd trust`/`qmd status` against a hostile-clone fixture: outside path not indexed unattended, indexed after `qmd trust`, re-armed when the path changes, in-project `./docs` still indexes without trust, `QMD_TRUST_UPDATE_HOOKS=1` opts in, custom model URIs skipped until trusted.

Existing #886 hook tests still pass. `tsc -p tsconfig.build.json --noEmit` is clean.

## Audit

Nested `picomatch` min-bumps for high GHSA-c2c7-rcm5-vvqj / GHSA-3v7f-55p6-f55p: micromatch 2.3.1 → 2.3.2, vite/tinyglobby 4.0.3 → 4.0.5. Direct picomatch stays 4.0.5. No zod/vitest majors. `flake.nix` not edited — Nix FOD hashes are stale after the lockfile change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,17 @@
   `~/.config/qmd/*.yml`, including anything `qmd collection update-cmd` writes,
   are unaffected.
 
+- The same project-local trust gate now covers collection `path` values that
+  resolve outside the project and `models.*` URIs that are not the built-in
+  defaults (#889). A checked-in `.qmd/index.yml` could previously point
+  `qmd update` at any readable directory and `qmd embed` / `qmd query` /
+  `qmd pull` at any `hf:` repo or local GGUF, with no prompt. In-project paths
+  (`./docs`) and default model URIs still apply without approval. Untrusted
+  out-of-project collections are skipped (other collections still index);
+  untrusted model URIs fall back to the built-in defaults or `QMD_*_MODEL`.
+  Approvals reuse `<config dir>/trusted.json`; a hooks-only digest is unchanged
+  so existing `qmd trust` records still match.
+
 - Indexing no longer follows file symlinks or glob `../` / absolute patterns
   out of the collection directory. `fast-glob` already skipped symlinked
   directories, but a file symlink (or a mask like `../**/*.md`) still resolved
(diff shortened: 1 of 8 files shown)


tobi:
Leftover twin of #892, which already landed on main (`40fb36f`) and closes #889. Closing so this does not sit open.

Issue #724, opened 2026-06-11
qmd embed cannot complete large collections in one run — session maxDuration (10 min) aborts mid-run with mass "LLM session expired before embedding chunk" failures


eohjun:
## Environment
- qmd 2.5.3 (`@tobilu/qmd` via npm, latest as of 2026-06-11)
- macOS 26.5.1, Apple M5 Pro (arm64), Node v26.3.0
- node-llama-cpp 3.18.1 (bundled; also reproduced with a local source build, llama.cpp b8390)
- Embedding model: `QMD_EMBED_MODEL=hf:Qwen/Qwen3-Embedding-0.6B-GGUF/Qwen3-Embedding-0.6B-Q8_0.gguf`
- GPU: Metal active (the `ggml_metal_library_init_from_source` / `tensor API not
  supported` warnings appear but classic Metal kernels work fine)

## Summary
`qmd embed` creates one LLM session for the whole run, and that session has a hard
default `maxDuration` of 10 minutes. On a collection large enough that embedding
takes longer than that, the session expires mid-run; the embed loop then records
every remaining chunk as failed and aborts instead of recreating the session and
continuing. A large collection can therefore never finish in a single run — only
repeated re-invocations of `qmd embed` slowly converge (progress IS committed
incrementally, which is great).

## Reproduction
1. A markdown collection big enough that embedding takes >10 min. Ours: ~1,250 md
   files (PKM vault with long notes; ~500 docs pending → tens of thousands of
   900-token chunks).
2. `qmd embed -c <collection>`
3. After ~10 minutes, stderr floods with one line per remaining chunk:
   ```
   <path>.md#<seq> (1 attempts): LLM session expired before embedding chunk
   ...and 4,009 more
   ```
   and the run exits. Pending count drops only by the docs fully completed before
   the cutoff (e.g. 503 → 457 in one run, → 428 the next, ...).

## Where it happens (dist of 2.5.3; line numbers from the published build)
- `dist/llm.js` ~1367: `const maxDuration = options.maxDuration ?? 10 * 60 * 1000; // Default 10 minutes`
  — a timer calls `abortController.abort(...)`, flipping `session.isValid`.
- `dist/store.js` ~1299–1306: the embed batch loop checks `if (!session.isValid)`
  and then `recordFailure(chunk, "LLM session expired before embedding chunk")`
  for ALL remaining chunks and breaks (comment: "Abort early if session has been
  invalidated (e.g. max duration exceeded)").
[...]

Issue #637, opened 2026-05-09
qmd embed silently leaves docs partially embedded after session timeout


tomasruta:
When `qmd embed` runs long enough to hit the 10-minute LLM session limit (default in `dist/llm.js:1049`), partially-processed documents are silently marked complete and never re-attempted on subsequent `qmd embed` runs.

## Reproduction

1. Embed a corpus large enough that the run exceeds 10 minutes (e.g. ~350 docs of mixed sizes).
2. Notice `Session expired — skipping N remaining chunks` warnings during the run.
3. After the run, `qmd status` reports `Pending: 0` even though some chunks were skipped.
4. Subsequent `qmd embed` invocations make no progress — the partially-embedded docs aren't picked up.

## Expected

Docs with skipped chunks should remain pending so the next `qmd embed` can re-process them.

## Actual

`getHashesNeedingEmbedding` in `dist/store.js` only checks for the existence of `seq=0`:

```sql
SELECT COUNT(DISTINCT d.hash) as count
FROM documents d
LEFT JOIN content_vectors v ON d.hash = v.hash AND v.seq = 0
WHERE d.active = 1 AND v.hash IS NULL
```

If chunk 0 succeeded but chunks 1..N failed, the doc is treated as fully embedded.

## Concrete impact

A 472-page book in my corpus ended up with 96 chunks indexed instead of the expected ~408 (about 75% of its content silently missing from search). The only signal was an anomalous KB-per-chunk ratio (7.6 vs ~1.8 for comparable books).

## Suggested fixes (in order of complexity)

1. **Cheap detection:** track expected chunk count per doc at chunking time and compare to actual `COUNT(*) FROM content_vectors WHERE hash = ?` to detect partial coverage.
2. **Robust:** on session expiry, mark the in-flight doc's hash for re-processing (e.g. delete its existing partial chunks so the next run starts fresh).
3. **Operator visibility:** at the end of `qmd embed`, surface a summary of any docs with `chunks < expected_chunks` so users notice partial failures.

## Workaround for current users

```sql
-- Identify partially-embedded docs (kb-per-chunk ratio significantly above norm)
SELECT d.collection, d.path,
       length((SELECT doc FROM content WHERE hash = d.hash))/1024 AS kb,
       (SELECT COUNT(*) FROM content_vectors WHERE hash = d.hash) AS chunks
FROM documents d
WHERE d.active = 1
  AND length((SELECT doc FROM content WHERE hash = d.hash)) /
[...]

PR #901, opened 2026-08-16
fix(embed): report actual completion and session expiry (additive)

KeilerHirsch:
## What

Additive, non-breaking reporting fix for the embed failure mode described in #724. `EmbedResult` now distinguishes between documents **selected** at run start (`docsProcessed`, kept for compatibility) and documents actually **completed** before the run ended (`docsCompleted`, `remainingDocs`, `sessionExpired`).

## Why

`generateEmbeddings()` returned `docsProcessed: totalDocs` — the pending selection at run start, not the number of documents completed before session expiry. The CLI printed `✓ Done! Embedded X chunks from Y documents` with Y = that selection and exited 0. A run stopped at the session cap can therefore look like a completed run to automation even though `qmd status` still shows pending documents.

The on-disk partial-state problem from #637 is already handled (keep-partial fix); this is specifically about **result semantics / operator visibility**.

## Changes

- `EmbedResult`: adds `docsCompleted`, `remainingDocs`, `sessionExpired`; `docsProcessed` retained and documented as "pending selection at run start".
- CLI summary: prints completed documents; when `remainingDocs > 0`, adds `⚠ N documents remaining (session expired at the duration cap | run stopped early) — re-run qmd embed to continue`. Exit code unchanged.
- Tests: extended the existing `maxDurationMs` expiry test and the full-run test to assert the new fields.

## Testing

- `npm run test:types` passes.
- vitest "Embedding batching" (`test/store.test.ts`): 14/14 pass, including the two extended tests.

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -2241,7 +2241,11 @@ async function vectorIndex(
       console.log(`${c.green}✓ No non-empty documents to embed.${c.reset}`);
     } else {
       console.log(`\r${c.green}${renderProgressBar(100)}${c.reset} ${c.bold}100%${c.reset}                                    `);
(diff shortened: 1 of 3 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #769, opened 2026-07-14
Add remote inference support: embeddings, query expansion, and reranking via Ollama + HTTP services

Fryerchristopher:
# Remote inference for QMD on low-power devices: embeddings, query expansion, and reranking via Ollama + HTTP services

## Summary

QMD's inference pipeline — embeddings, query expansion, reranking, tokenization, and LLM session management — relies entirely on local `llama.cpp`. On resource-constrained devices (e.g., Raspberry Pi 5), local inference is too slow for practical use: 87 seconds per vector search and 25+ minutes for a full re-embed of ~800 documents.

This PR adds support for **remote inference** via three environment variables, allowing QMD to offload all LLM operations to a more powerful machine on the network (e.g., a Mac Studio with Ollama, a GPU workstation, or a cloud instance):

| Variable | Purpose | Default |
|----------|---------|---------|
| `QMD_OLLAMA_EMBED_URL` | Remote Ollama for embeddings (`/api/embed`) | unset (local llama.cpp) |
| `QMD_OLLAMA_EMBED_MODEL` | Embedding model name in Ollama | `nomic-embed-text` |
| `QMD_EXPAND_URL` | Remote Ollama for query expansion (`/api/generate`) | unset (local or skip) |
| `QMD_EXPAND_MODEL` | Generation model for expansion | `qwen3:0.6b` |
| `QMD_RERANK_URL` | Remote HTTP reranker service (`/rerank` endpoint) | unset (local or skip) |

When none are set, QMD behaves exactly as before — **full backward compatibility**.

When `QMD_OLLAMA_EMBED_URL` is set, the three remaining local-LLM operations (query expansion, reranking, tokenizer-based chunking, and LLM session initialization) are either routed to remote services (`QMD_EXPAND_URL`, `QMD_RERANK_URL`) or skipped with graceful fallback (char-based chunking, raw query). This gives users a choice: minimal setup (embeddings only, skip expansion/reranking) or full quality (all three services running on the remote host).

This PR also includes a **reference reranker service** (`contrib/reranker-service.py`) — a standalone Python/Flask app that loads the Qwen3-Reranker-0.6B cross-encoder model via PyTorch and serves the `/rerank` endpoint that `QMD_RERANK_URL` calls.

### Performance results

| Metric | Before (local llama.cpp) | After (all remote) | Improvement |
|--------|--------------------------|---------------------|------------|
| Vector search latency | 87s | 0.12s | **725x** |
[...]

Issue #620, opened 2026-05-02
Support OpenAI-compatible backends for generation, embeddings, and reranking


loopyd:
# Problem

QMD is currently centered around local `llama.cpp` model access.

That makes it harder to use QMD with a local or self-hosted OpenAI-compatible server when a user wants to route:
- query expansion / generation
- embeddings
- reranking

through the same remote service.

There are already requests for remote embeddings, but the broader problem is allowing QMD to work with a compatible server across the whole retrieval flow.

Related issues:
- #403
- #428
- #521

# Solution

Add support for an OpenAI-compatible backend that can be configured in QMD and used for:
- generation
- embeddings
- reranking

This should allow users to point QMD at a compatible local or self-hosted server, configure model aliases, and run the normal QMD flow without requiring direct local model access.

# Reference

PR #619 implements this feature:
https://github.com/tobi/qmd/pull/619

PR #902, opened 2026-08-16
feat: add fail-closed shared Qwen embedding and reranking

coolmanns:
## Summary

Adds strict, fail-closed HTTP backends for shared embedding and reranking services while preserving QMD's existing local model-string behavior.

This draft is intentionally narrower than the complete generation + embedding + reranking backend requested in #620. It contributes a hardened embedding/reranking path for a concrete Qwen3 deployment:

- OpenAI-compatible `POST /v1/embeddings`
- llama.cpp-compatible `POST /v1/rerank`
- native Qwen3 2,560-dimensional validation
- first-1,024 Matryoshka prefix reduction
- L2 normalization after truncation
- no local model initialization or fallback in remote mode

## Motivation

Multiple QMD clients should be able to reuse one embedding model and one reranking model without each process loading duplicate GGUF models.

The important correctness boundary is stronger than transport compatibility: vectors from different producers, preprocessing paths, dimensions, or normalization pipelines must never be silently mixed under one identity. Remote failure therefore fails closed rather than falling back to a local GGUF model.

## Configuration

```yaml
models:
  embed:
    provider: openai
    endpoint: http://127.0.0.1:8086/v1
    model: Qwen3-Embedding-4B-Q8_0.gguf
    nativeDimensions: 2560
    dimensions: 1024
    reduction: mrl-prefix
    normalization: l2
    formatVersion: qwen3-query-document-v1
  rerank:
    provider: openai
    endpoint: http://127.0.0.1:8088/v1
    model: Qwen3-Reranker-0.6B-Q4_K_M.gguf
    failurePolicy: fail-closed
```

Plain model strings remain local configurations, including strings that happen to look like HTTP URLs. Remote mode requires the explicit structured form.

## Embedding safety contract

- Supports single and batched input.
- Maps entries using each returned `index`, never response position.
- Rejects missing, duplicate, fractional, negative, and out-of-range indexes.
- Requires exactly 2,560 native values for this Qwen3 configuration.
- Validates every native value before truncation, including the discarded tail.
- Rejects non-finite values and finite JS doubles that overflow Float32.
- Takes exactly the first 1,024 dimensions, then L2-normalizes the prefix.
- Rejects zero/invalid prefix norms and invalid normalized Float32 values.
- Applies the identical transformation to document and query embeddings.
- Requires response model identity, with only explicit canonical aliases accepted.
- Uses a semantic fingerprint covering provider, model, native/output dimensions, reduction, normalization order, and formatting version; endpoint and credentials are excluded.
- Never initializes `node-llama-cpp` or silently falls back locally in remote mode.
- Stages remote results in memory before a bounded SQLite write, avoiding network awaits inside transactions/savepoints.
- Prevents overlapping remote embedding invocations against the same store.

## Reranking safety contract

- Sends chunks as `documents` and maps results by returned document index.
- Validates cardinality, unique/in-range integer indexes, finite normalized scores, and response model identity.
- Uses normalized `relevance_score` directly without applying a second sigmoid.
- Caches by returned document index rather than file path.
- Separates cache identity for different remote backends without including credentials.
- Uses explicit `failurePolicy: fail-closed` and never initializes/falls back to a local reranker in remote mode.

## CLI and configuration behavior

Structured configuration is validated before status, doctor, pull, trust, update, embed preflight, and runtime use. Malformed remote configuration does not degrade into a local path. Diagnostics redact API keys, URL credentials/query secrets, response bodies, and raw transport causes.

## Relation to existing work

This draft overlaps with and is informed by:

- #620 — broader OpenAI-compatible generation/embedding/reranking goal
- #705 — RemoteLLM/HybridLLM architecture
- #761 — remote OpenAI-compatible embedding backend
- #769 — remote inference for low-power clients

It specifically addresses correctness concerns raised during review of those PRs:

- fail closed rather than mix local/remote vector producers;
- validate response cardinality and complete returned-index coverage;
- reject malformed dimensions and non-finite values;
- fingerprint the actual semantic transform;
- separate reranking cache identities;
- add focused routing/config/store/CLI tests;
- avoid duplicate local model initialization.

This is submitted as a draft because the Qwen3 2,560→1,024 MRL contract is intentionally explicit and opinionated. Maintainer guidance is welcome on whether to retain this strict profile, generalize the structured transform fields, or extract parts into the broader #620 implementation.

## Verification

Focused local test lanes:

- embedding/reranking/routing/LLM/store: **452 passed, 0 failed**
- CLI/config/SDK/trust/update-hook: **318 passed, 0 failed**
- TypeScript check: passed
- build: passed
- `git diff --check`: passed

Full Vitest run:

- **1,351 passed**
- **1 failed**: `MCP HTTP Transport > POST /mcp initialize still works for 2025-era clients (no session id)`

That MCP failure was reproduced on the untouched base commit and is unrelated to this change; this PR does not claim a completely green full suite.

Additional isolated endpoint verification against llama.cpp services confirmed:

- adapter output length: 1,024
- post-prefix norm: approximately 1.0
- manual native-prefix transform cosine: approximately 1.0
- retrieval fixture: Recall@3 1.00, MRR 0.85
- bounded four-worker load: 20/20 requests successful
- live TCP traffic to both embedding and reranking endpoints
- no additional embedding or reranking model process loaded

No production database, generated index, credential, benchmark artifact, or deployment configuration is included in this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixes
+
+- Doctor: distinguish bounded remote-embedding batch-shape numerical drift from stale or corrupt vectors, while retaining strict local checks and fail-closed handling for material remote mismatches.
+
 ### Security
 
 - `qmd update` no longer runs a project-local `.qmd/index.yml`'s `update:`
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -9,6 +9,7 @@ import { basename, dirname, join as pathJoin, relative as relativePath, resolve 
 import { parseArgs } from "util";
 import { readFileSync, readdirSync, realpathSync, statSync, existsSync, unlinkSync, writeFileSync, openSync, closeSync, mkdirSync, lstatSync, rmSync, symlinkSync, readlinkSync, copyFileSync } from "fs";
 import { createInterface } from "readline/promises";
+import { createHash } from "node:crypto";
 import {
   getPwd,
   getRealPath,
@@ -86,7 +87,7 @@ import {
   type ReindexResult,
   type ChunkStrategy,
 } from "../store.js";
-import { disposeDefaultLlamaCpp, getDefaultLlamaCpp, setDefaultLlamaCpp, LlamaCpp, withLLMSession, pullModels, DEFAULT_MODEL_CACHE_DIR, resolveEmbedModel, resolveGenerateModel, resolveRerankModel, resolveModels, inspectGgufFile, isDarwinMetalMitigationActive } from "../llm.js";
+import { disposeDefaultLlamaCpp, getDefaultLlamaCpp, setDefaultLlamaCpp, LlamaCpp, withLLMSession, withLLMSessionForLlm, pullModels, DEFAULT_MODEL_CACHE_DIR, resolveEmbedModel, resolveGenerateModel, resolveRerankModel, resolveModels, inspectGgufFile, isDarwinMetalMitigationActive } from "../llm.js";
 import {
   formatSearchResults,
   formatDocuments,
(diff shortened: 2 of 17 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #903, opened 2026-08-16
fix(store): serialize concurrent FTS table create on cold open

tobi:
Fixes the Bun/macOS CI failure on main: `table documents_fts already exists` from `concurrent store initialization > cold database` in `test/store-concurrency.test.ts` ([run 31734426108](https://github.com/tobi/qmd/actions/runs/31734426108)).

## What was wrong

Trigger rebuild and CJK publish were already serialized with `BEGIN IMMEDIATE` + a double-checked read. The remaining hole was the FTS **table** itself.

`CREATE VIRTUAL TABLE IF NOT EXISTS documents_fts` ran as an autocommit statement in `initializeDatabase`. FTS5's `IF NOT EXISTS` is not atomic across WAL connections: two first-open processes can both observe a missing table on their schema snapshot, and the loser throws `table documents_fts already exists`. `recreateDocumentsFts` had the same check-then-`CREATE` window without `IF NOT EXISTS`. Bun/macOS SQLite locking makes that window show up in CI; Node and Linux usually miss it.

## The fix

- Move first-create and legacy-schema repair into `ensureDocumentsFtsSchema`, under the same `BEGIN IMMEDIATE` + double-check as the sync triggers.
- Keep `CREATE VIRTUAL TABLE IF NOT EXISTS` for the uncontended path.
- Treat a concurrent `already exists` as success when `documents_fts` is present.
- Stop autocommit-creating the FTS table in `initializeDatabase`.

## Tests

- Structural assertion that FTS DDL is `IF NOT EXISTS`, create/repair is inside `BEGIN IMMEDIATE`, and `initializeDatabase` no longer autocommit-creates the table.
- Cold-open multi-process test now runs two trials (the single-schedule flake that misses the race).
- Local: bun `test/store-concurrency.test.ts` + `test/store-cjk-fts.test.ts` + store slice + `test/store.test.ts` green. Node 22 vitest concurrency + CJK + store slice green. No release. `flake.nix` / `flake.lock` untouched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,6 +81,15 @@
 
 ### Fixed
 
+- Concurrent first-open of a cold index no longer fails with
+  `table documents_fts already exists` on Bun/macOS. FTS5
+  `CREATE VIRTUAL TABLE IF NOT EXISTS` is not atomic across WAL
+  connections: two processes can both see a missing table on their
+  schema snapshot and the loser throws. Table create and legacy-schema
+  repair now use the same `BEGIN IMMEDIATE` + double-check as the FTS
+  sync triggers, and treat a concurrent "already exists" as success
+  when the table is present.
+
 - Nix flake `qmd-node-modules` FOD hashes updated for x86_64-linux and
   aarch64-darwin after the MCP SDK 2.0 bump. `nix build` / Nix GHA was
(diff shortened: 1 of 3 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #904, opened 2026-08-17
fix(cli): bound doctor embedding sample query

naveenspark:
## Summary

On large indexes, `qmd doctor` consumes excessive memory because its embedding health check joins full document bodies before applying `ORDER BY random() LIMIT 3`. SQLite can therefore materialize and sort the indexed corpus just to select three samples.

This changes the query to:

1. Select three `(hash, seq)` identifiers first.
2. Join document bodies only for those selected samples.

Active-document filtering and minimum-path selection remain unchanged.

## Reproduction

Reproduced on a multi-gigabyte local index with tens of thousands of documents:

- Before: process memory grew into the tens of gigabytes and the command did not complete.
- After: `qmd doctor` completes in a few seconds.

## Validation

- Added cross-runtime coverage for the existing selection semantics.
- Added a regression test proving document-body reads are bounded to the requested sample size.
- `bun run test:unit`
- `bun run test:types`
- `bun run build`
- `bun run test:package`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `qmd doctor` now chooses embedding sample identifiers before joining full
+  document bodies. This prevents its random sample query from materializing
+  and sorting the indexed corpus, which could exhaust memory on large indexes.
+
 ## [2.8.3] - 2026-08-16
 
 ### Security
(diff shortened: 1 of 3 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

naveenspark:
Closing this stale/conflicted lane during local worktree cleanup. The remote branch remains recoverable. Any renewed doctor work will start from current upstream with a focused reproduction and regression for the remaining shutdown behavior.

Issue #803, opened 2026-07-29
vsearch -c <collection> returns nothing for small collections (filter applied after global ranking)


Tooke:
`vsearch -c <collection>` returns nothing for small collections, because `-c` filters a globally-ranked pool instead of scoping the query.

**Version:** qmd 2.5.3 (b48c7994b5), macOS / Homebrew install of `@tobilu/qmd`.

## Summary

On the vector path, `-c` appears to be applied *after* ranking, to a global candidate pool whose size derives from `-n`. A collection that is small relative to the index therefore never reaches the pool, and the filter returns an empty result — even though the collection is fully indexed and embedded, and its documents rank fine in an unfiltered search.

The lexical path (`qmd search -c ...`) behaves correctly. Only `vsearch` is affected.

## Reproduction

An index with six collections; the two large ones are ~4,200 and ~140 files, the rest are 4–37 files. Querying a 5-file collection:

```
$ qmd vsearch "<query>" -c small-collection -n 3   --json | grep -c '"file"'
0
$ qmd vsearch "<query>" -c small-collection -n 20  --json | grep -c '"file"'
2
$ qmd vsearch "<query>" -c small-collection -n 100 --json | grep -c '"file"'
2
$ qmd vsearch "<query>" -c small-collection -n 500 --json | grep -c '"file"'
3
```

The collection contains only 5 documents, so `-n 3` should be able to return up to 3 of them. Instead it returns 0, and hits only start appearing once `-n` is large enough that the global pool happens to reach into that collection.

The same query **unfiltered** returns documents from that collection without difficulty:

```
$ qmd vsearch "<targeted query>" -n 25 --json | jq -r '.[].file' | grep '^qmd://small-collection/'
qmd://small-collection/<doc>.md
```

So the documents are reachable, embedded, and competitive — they are just being filtered out of a pool they were never allowed into.

## Ruled out

- **Missing embeddings.** `qmd embed -f -c small-collection` re-embeds cleanly (44 chunks from 5 documents in 2s) and changes nothing. `qmd status` reports the collection as indexed and current.
- **Collection naming / registration.** `qmd ls small-collection` lists the files, and `qmd search -c small-collection` (lexical) returns hits for the same query immediately.

## Scaling of the symptom

In the same index, per-collection `vsearch -c … -n 3`:

| collection size | hits |
|---|---|
| ~4,200 files | 3 |
| ~140 files | 3 |
| 37 files | 0 |
| 35 files | 0 |
[...]

Issue #791, opened 2026-07-25
bug: Vector search returns nothing for small collections: global ANN candidates are filtered *after* retrieval


01Gibbs:
# Vector search returns nothing for small collections: global ANN candidates are filtered *after* retrieval

**Version:** @tobilu/qmd 2.5.3 (Homebrew, Linux/WSL2)

## Summary

`searchVec` takes a fixed-size **global** nearest-neighbour candidate set and only
then filters by collection. When the requested collection is a small share of the
vector index, none of the global candidates belong to it, and the search returns
zero results — even for queries that obviously match documents in that collection.

`searchFTS` already compensates for exactly this problem; `searchVec` does not.

## The asymmetry

`dist/store.js:2614` — lexical path widens the candidate pool when filtering:

```js
const ftsLimit = collectionName ? limit * 10 : limit;
```

`dist/store.js:2685` — vector path does not:

```js
).all(new Float32Array(embedding), limit * 3);
```

Step 1 is a global `MATCH ... AND k = ?` over `vectors_vec` with no collection
predicate (correctly so — sqlite-vec can't join). Step 2 applies
`AND d.collection = ?` to whatever those `limit * 3` rows happened to be. With
`limit = 10` that is 30 global candidates, and if the target collection is a small
fraction of the index, the expected number surviving the filter is < 1.

## Reproduction

Index with two collections in one DB:

| collection | vectors | share |
|---|---:|---:|
| `docling-chunks` | 24,904 | 37.8% |
| `wiki` | 451 | 0.7% |
| (total index) | 65,853 | |

```
$ qmd vsearch "tikanga Māori legal order" -c wiki
No results found.

$ qmd vsearch "tikanga Māori legal order" -c docling-chunks
5 results

$ qmd search "tikanga" -c wiki
5 results          # lexical is fine — it fetches limit*10
```

`wiki` is fully and freshly embedded — 451 vectors, same model and
`embed_fingerprint` (`c37385`) as the large collection, so this is not staleness:

```sql
SELECT model, embed_fingerprint, COUNT(*) FROM content_vectors cv
JOIN documents d ON d.hash = cv.hash WHERE d.active=1 AND d.collection='wiki'
GROUP BY 1,2;
-- hf:ggml-org/embeddinggemma-300M-GGUF/... | c37385 | 451
```

Confirming it's crowding rather than relevance: the same query run **unscoped**
returns no `wiki` page anywhere in the global top-20 — the wiki pages are
[...]

PR #910, opened 2026-08-19
Add metadata support

aaronccasanova:
Documents can opt into typed metadata through frontmatter, and every search surface can filter on it.

```markdown
---
qmd:
  metadata:
    topics: [typescript, programming]
    status: published
    priority: 3
---

# Document body starts here
```

```sh
qmd query "dependency injection" --filter '{"key":"status","operator":"eq","value":"published"}'
```

The same filter works on `search`, `vsearch`, and `query`, on the SDK's `search()`/`searchLex()`/`searchVector()`, on the MCP `query` tool, and on HTTP `POST /query`. Results carry the document's metadata in JSON, SDK, MCP, and HTTP output. Documents without `qmd.metadata` behave exactly as before, and unfiltered search benchmarks unchanged.

The motivating use case is agentic retrieval. Relevance search answers "what is about X", but agents also need structured slicing: only published docs, only a given audience, only priority above a threshold. Today that means over-fetching and post-filtering client-side, with no way to express "and don't show me drafts" at all.

Collections already scope retrieval, and this composes with them rather than replacing them. A collection is one coarse dimension, where a document lives, chosen at indexing time. Metadata lets each document declare its own dimensions (status, audience, priority) and lets every query combine them, so `-c notes --filter ...` narrows by place and by properties at once.

### What this adds

- An opt-in, namespaced `qmd.metadata` frontmatter block (`.md`, `.markdown`, `.mdx`) extracted at index time. Values are strings, numbers, booleans, or flat homogeneous arrays of one of those.
- One recursive filter language, shared verbatim by CLI, SDK, MCP, and HTTP.
- Two derived SQLite tables holding canonical metadata JSON plus typed, indexed rows for filtering.
- `qmd status` reports documents pending extraction, and a plain `qmd update` backfills existing indexes (no schema version bump, no forced re-embedding).

### What this deliberately does not change

- Raw content is untouched. Frontmatter stays ordinary searchable text: same content hash, same FTS body, same chunking and embeddings, same snippets and line numbers. #770's evaluation showed no retrieval win from stripping frontmatter, so this PR does not touch that path.
- No tagging subsystem. `tags` is an ordinary user-defined key with no special behavior. Multi-valued classification is array-valued metadata.
- No index-only mutation API. The SQLite tables are a derived index, and the document remains the authority, so metadata survives rebuilds, syncs across machines, and reviews in version control.
- No new dependencies. Extraction reuses the existing `yaml` package.
- Nothing changes for anyone who never writes `qmd.metadata`.

### The filter language

A filter is a JSON AST discriminated by `operator`:

| Node | Shape |
|------|-------|
| Logical group | `{ "operator": "and" \| "or", "operands": [...] }` |
| Negation | `{ "operator": "not", "operand": {...} }` |
| Comparison | `{ "key", "operator": "eq" \| "ne" \| "gt" \| "gte" \| "lt" \| "lte", "value" }` |
| Membership | `{ "key", "operator": "in" \| "nin" \| "all", "value": [...] }` |
| Presence | `{ "key", "operator": "exists", "value": true \| false }` |

```sh
qmd query "dependency injection" --filter '{
  "operator": "and",
  "operands": [
    { "key": "topics", "operator": "all", "value": ["typescript", "programming"] },
    { "key": "status", "operator": "nin", "value": ["draft", "archived"] },
    { "operator": "or", "operands": [
      { "key": "priority", "operator": "gte", "value": 3 },
      { "key": "reviewed", "operator": "eq", "value": true }
    ] }
  ]
}'
```

Semantics are strict and typed: no coercion, type mismatches never match (including `ne` and `nin`), arrays match when any element satisfies a condition (`all` requires every filter value), and missing keys only match through `exists: false`. Composition is always explicit: no implicit AND, no `$`-prefixed shorthand, no reserved metadata keys. Validation errors point at the failing node by JSON path.

Two guarantees:

- Soundness is absolute. Every returned result satisfies the filter, applied before RRF fusion and reranking on every path.
- Completeness is best-effort for top-K under highly selective filters, matching the existing collection-filter contract exactly (candidate over-fetch, then relational filtering).

### Design notes

The shape of the filter language came out of a survey of metadata filtering in Pinecone, Chroma, MongoDB, Weaviate, Qdrant, and Elasticsearch, and the storage and retrieval design generalizes machinery already in QMD rather than adding parallel systems. Details collapsed below.

<details>
<summary>Why an explicit operator-discriminated AST (and not a $-prefixed DSL)</summary>

The operator vocabulary is the established one: `and`/`or`/`not` (logical), `eq`/`ne`/`gt`/`gte`/`lt`/`lte` (comparison), `in`/`nin`/`all` (membership), and `exists` (presence) all carry their MongoDB/Pinecone/Chroma meanings. `key` for the metadata key follows Qdrant. The node shape (one `operator` property discriminating groups with `operands` from leaf conditions) follows Weaviate's public GraphQL filter API.

The deliberate departure is rejecting the compact Mongo-style form (`{ status: "published", $or: [...] }`). That shape cannot be expressed precisely in TypeScript: an index signature cannot say "every string except the reserved `$` keys", so arbitrary keys broaden into a union of logical and field shapes, excess-property checking degrades, and the SDK type stops matching the runtime grammar. With one finite `operator` discriminator:

- The SDK exports a discriminated union that narrows per operator, so operator-specific `value` types are compile-time checked.
- The runtime validator mirrors the same union and rejects unknown properties, with errors that name the failing JSON path.
- Metadata keys stay pure data and can never collide with grammar keys.

`operands`/`operand` (over `filters`/`filter` or `rules`) is the standard term for the recursive expressions a logical operator consumes, and the singular form encodes `not`'s unary arity in the type system. Leaf conditions keep explicit `key` and `value` roles instead of a positional array.

Validation enforces bounded depth, operand counts, key lengths, and value sizes (shared limits across all four surfaces).

</details>

<details>
<summary>Storage: two derived tables, extraction state, typed rows</summary>

```sql
document_metadata          -- one row per document
  document_id PRIMARY KEY REFERENCES documents(id) ON DELETE CASCADE
  metadata_json            -- canonical JSON, returned with results
  extraction_version       -- current extractor version (backfill trigger)
  extraction_error         -- bounded message when frontmatter is invalid

document_metadata_values   -- one row per (key, element), typed
  (document_id, key, ordinal) PRIMARY KEY
  value_type + text_value / number_value / boolean_value
  CHECK: exactly one value column populated, matching value_type
```

Three covering partial indexes (`(key, text_value, document_id)` and the number/boolean equivalents) serve the filter compiler. Metadata attaches to documents rather than content hashes because two documents can share a hash while carrying different metadata, and filters conceptually select documents.

Metadata is data, not schema, so there is no `user_version` bump: the tables are `CREATE TABLE IF NOT EXISTS` on open, existing indexes keep working, and `qmd update` backfills extraction for unchanged documents (tracked by `extraction_version`). Invalid metadata records a bounded per-document error, the document still indexes and searches normally, and `qmd update` surfaces a warning. `qmd status` shows the pending count.

Extraction is defensive: BOM/CRLF tolerated, a missing closing delimiter means "no frontmatter" rather than an error, YAML alias expansion is bounded, nested objects/nulls/mixed arrays are rejected with precise messages, and arrays are deduplicated preserving order.

</details>

<details>
<summary>Retrieval integration and the completeness contract</summary>

The compiler turns a filter into parameterized correlated `EXISTS`/`NOT EXISTS` subqueries over `document_metadata_values` (all user data bound, never interpolated). One compiler serves both retrieval backends, so FTS and vector search cannot drift semantically.

FTS keeps the existing CTE that forces FTS5 to run first (the comment in `searchFTS` documents the full-scan regression this prevents), then applies the compiled predicate alongside the collection filter with the same `limit * 10` candidate over-fetch. Filtered search additionally requires current, error-free extraction, so a document that has not been processed cannot accidentally satisfy `exists: false`.

Vector search generalizes the collection exact-scan path from #791/#803: a filter resolves an eligible document set, eligible sets up to 20,000 vectors get an exact cosine scan (no post-filter starvation of small sets), larger sets fall back to global ANN with the capped over-fetch. The filter is re-applied on the document join because vectors are content-scoped while metadata is document-scoped, which also handles shared-hash documents correctly.

Hybrid `query` passes the filter to both backends before RRF and reranking. Results attach metadata via one batch lookup (LEFT JOIN in the backends, one query in hybrid paths), no per-row N+1.

sqlite-vec 0.1.9's native vec0 metadata columns were evaluated and don't fit the general case (16 fixed columns, no multi-valued arrays, dynamic user-defined keys), though a native pre-filter through one internal id column is a plausible follow-up optimization. Pinned docs: [vec0](https://github.com/asg017/sqlite-vec/blob/v0.1.9/site/features/vec0.md), [knn](https://github.com/asg017/sqlite-vec/blob/v0.1.9/site/features/knn.md).

</details>

### Performance

Benchmarked at 10,000 documents (Apple M3 Pro, fresh database per scenario, p50 of 300 timed iterations after warmup), baseline v2.8.3 vs this branch:

| Scenario | Baseline | This branch |
|---|---|---|
| Unfiltered FTS (query matching all 10k docs) | 8.07ms | 8.08ms (no metadata) / 8.15ms (metadata on all docs) |
| Unfiltered vector KNN | 14.7ms | 14.7ms / 13.7ms |
| Filtered FTS, any selectivity from 90% to 0% | n/a | 7.9-8.6ms |
| Filtered vector, 90% eligible (exact scan) | 306ms (collection scope, same path) | 323ms |

Unfiltered search is unchanged within noise. Filtered FTS costs at most ~0.5ms over unfiltered, even for nested filters. Filtered vector search inherits the existing collection exact-scan cost profile (~5% over collection scoping at equal eligible-set size, for the EXISTS predicates), and the collection path itself does not regress.

Scripts and full results are reproducible from [`benchmarks/` on a companion branch](https://github.com/aaronccasanova/qmd/tree/bench/metadata-filtering/benchmarks) of my fork, kept off this branch to keep the diff reviewable.

<details>
<summary>Query plans</summary>

Unfiltered `searchFTS` keeps the FTS-first CTE shape. The new `LEFT JOIN document_metadata` (which returns result metadata without an N+1) is a rowid primary-key lookup on only the emitted rows, which is why the unfiltered delta doesn't show up in measurement:

```text
CO-ROUTINE fts_matches
SCAN documents_fts VIRTUAL TABLE INDEX 0:M3
USE TEMP B-TREE FOR ORDER BY
SCAN fm
SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
SEARCH content USING INDEX sqlite_autoindex_content_1 (hash=?)
SEARCH dm USING INTEGER PRIMARY KEY (rowid=?) LEFT-JOIN
```

Filtered `searchFTS` (`and(status eq, priority gte)`) keeps the same shape and resolves every EXISTS through the covering partial indexes:

```text
CO-ROUTINE fts_matches
SCAN documents_fts VIRTUAL TABLE INDEX 0:M3
USE TEMP B-TREE FOR ORDER BY
SCAN fm
SEARCH dm USING INTEGER PRIMARY KEY (rowid=?)
SEARCH d USING INTEGER PRIMARY KEY (rowid=?)
SEARCH content USING INDEX sqlite_autoindex_content_1 (hash=?)
SEARCH mv EXISTS USING COVERING INDEX idx_metadata_text_lookup (key=? AND text_value=? AND document_id=?)
SEARCH mv EXISTS USING COVERING INDEX idx_metadata_number_lookup (key=? AND number_value>?)
```

</details>

### Testing

72 new tests across six suites, green under both Node (vitest) and Bun:

- Extraction: frontmatter edge cases (BOM, CRLF, missing delimiters, alias bombs, every rejection path)
- Filter language: validation with JSON-path errors, SQL semantics against a real SQLite database, and SQL-injection payloads staying inert data
- Persistence: replacement, shared content hashes, cascade deletes, and `qmd update` backfill
- Search integration: FTS/vector/structured with precomputed embeddings (no model downloads in CI)
- Surfaces: SDK, MCP tool, HTTP endpoints, and spawned-CLI runs including exit codes and error messages

Full suite: typecheck, 1215 Node tests, 1215 Bun tests, and the package smoke all pass. Also verified end to end with real models (embedding, expansion, reranking) against a live index.

### Review guide

The five commits build the feature in dependency order, each independently green:

1. `feat(metadata): add metadata domain modules` (extraction, filter AST + validation + SQL compiler, storage, pure and heavily tested)
2. `feat(metadata): persist document metadata during indexing`
3. `feat(metadata): filter FTS, vector, hybrid, and structured search`
4. `feat(metadata): expose --filter and filter options on every search surface`
5. `docs(metadata): document metadata and metadata filtering`

### Limitations and future work

- Filtered vector search over large eligible sets pays the exact-scan cost (same as collection scoping today). A native sqlite-vec pre-filter through an internal id metadata column is a candidate follow-up.
- Conditions match array elements independently. There is no `elemMatch`-style "same element satisfies all" operator yet.
- Strings compare exactly and case-sensitively, dates stay strings, and dots in keys carry no path semantics. Each is a deliberate first-version restriction that can be relaxed compatibly.
- Frontmatter is the only metadata source. The extraction layer produces a canonical object, so external sources (config files, sidecars) can be added without touching the filter or storage layers.

### Related work

- #551 and #770: prior frontmatter work. This PR follows #770's evidence and leaves content processing alone.
- #703 and #441: earlier filtering attempts across surfaces, useful context for why one shared AST beats per-surface flags.
- #791 and #803: the collection-starvation fixes whose exact-scan machinery the filtered vector path generalizes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added Oxlint lint fence.
+- Document metadata and metadata filtering. Markdown documents can opt into typed metadata through a namespaced frontmatter block (`qmd.metadata` with strings, numbers, booleans, or flat homogeneous arrays), and every search surface — CLI `search`/`vsearch`/`query` via `--filter <json>`, the SDK's `filter` option on `search()`/`searchLex()`/`searchVector()`, the MCP `query` tool, and HTTP `POST /query` and `/search` — accepts one shared recursive filter AST discriminated by `operator`: `and`/`or`/`not` logical groups, `eq`/`ne`/`gt`/`gte`/`lt`/`lte` comparisons, `in`/`nin`/`all` membership, and `exists` presence. Every returned result satisfies the filter (applied before RRF fusion and reranking); like collection filtering, highly selective filters remain best-effort for top-K completeness. Frontmatter stays ordinary searchable content — no chunking, embedding, snippet, or line-number changes — and documents without `qmd.metadata` behave exactly as before. JSON/SDK/MCP/HTTP results now include each document's indexed metadata, and `qmd status` reports how many documents still need metadata extraction (a normal `qmd update` backfills existing indexes).
 
 ## [2.8.3] - 2026-08-16
 
--- a/README.md
+++ b/README.md
@@ -151,7 +151,7 @@ runs in a container and a liveness probe connects from a non-loopback address.
 
 The HTTP server exposes two endpoints:
 - `POST /mcp` — MCP Streamable HTTP (JSON responses, stateless)
-- `POST /query` (alias `/search`) — structured search without the MCP protocol
+- `POST /query` (alias `/search`) — structured search without the MCP protocol. Accepts the same optional `filter` object as the `query` tool (invalid filters return `400`); see [Metadata Filtering](#metadata-filtering)
 - `GET /health` — liveness check with uptime
 
 
@@ -187,6 +187,7 @@ Point any MCP client at `http://localhost:8181/mcp` to connect.
 |------|-----------|------|-------|
 | `query` | `searches` | array | Typed sub-queries (`lex`/`vec`/`hyde`), 1–10. **Required.** First gets 2x weight. |
 | `query` | `collections` | string[] | Filter by collection names (OR). **Array only** — singular `collection` is silently ignored. |
+| `query` | `filter` | object | Metadata filter (recursive `operator`-discriminated JSON AST; see [Metadata Filtering](#metadata-filtering)) |
 | `query` | `intent` | string | Disambiguation context (does not search on its own) |
 | `query` | `limit` | number | Max results (default 10) |
 | `query` | `minScore` | number | Minimum relevance 0–1 (default 0) |
@@ -292,6 +293,20 @@ const results3 = await store.search({
 
 // Skip reranking for faster results
 const fast = await store.search({ query: "auth", rerank: false })
+
+// Metadata filter — every returned result satisfies it (also available on
+// searchLex() and searchVector()); results expose indexed metadata via
+// r.metadata. See "Metadata Filtering" for the full grammar.
+const published = await store.search({
+  query: "authentication flow",
+  filter: {
+    operator: "and",
+    operands: [
+      { key: "topics", operator: "all", value: ["typescript"] },
+      { key: "status", operator: "ne", value: "draft" },
+    ],
+  },
+})
 ```
 
 For direct backend access:
@@ -860,6 +875,7 @@ and `deep-search` (→ `query`).
 --full             # Show full document content
 --line-numbers     # Add line numbers to output
 --explain          # Include retrieval score traces (query, JSON/CLI output)
+--filter <json>    # Metadata filter (recursive JSON AST; see Metadata Filtering)
 --index <name>     # Use named index
 --intent "<text>"  # Disambiguation context (e.g. "web page load times")
 --no-rerank        # Skip LLM reranking (RRF scores only; faster on CPU)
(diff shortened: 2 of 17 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #912, opened 2026-08-19
fix(runtime): bound native shutdown and centralize Metal policy

rszrszrsz:
## Summary

- coordinate CLI, stdio MCP, and HTTP MCP teardown through one shutdown owner
- give every launch path a bounded graceful shutdown, with supervised first/second-signal behavior
- centralize Darwin Metal residency policy at every `getLlama()` attempt
- prevent direct LLM operations from entering while idle maintenance disposes native contexts
- preserve the exact non-zero exit code of failed collection update hooks

## Why

The existing lifecycle behavior depended on `bin/qmd` being present. Direct Node/Bun runs, Nix wrappers, and detached HTTP daemons could wait forever during shutdown. The wrapper could also deliver one terminal signal twice to its child, turning a normal drain into an immediate hard stop.

Metal residency policy was split between launchers and contradicted the Nix path. Direct embedding/reranking operations could race idle disposal because they did not wait on the maintenance barrier.

This PR keeps the existing runtime structure and fixes those ownership boundaries without adding another supervisor or process protocol. The wrapper sets one bit, `QMD_SUPERVISED=1`; detached daemons explicitly remove it.

## Metal decision

The default follows node-llama-cpp 3.20: residency sets disabled. `QMD_METAL_KEEP_RESIDENCY=1` is an explicit opt-in, while any explicitly set `GGML_METAL_NO_RESIDENCY` remains authoritative.

On Apple M1 Pro / macOS 14.7.6, an out-of-process A/B covered residency kept/default/explicitly disabled, with and without disposal:

- 30 embedding probe runs
- 18 heavier embedding + generation runs
- Metal selected in all relevant cells
- 48/48 model operations and exits succeeded
- no assertion, abort, segmentation fault, or teardown crash

## Verification

Before rebasing onto the latest `main`:

- full Vitest: 1,224 passed, 78 skipped
- full Bun: 1,224 passed, 87 skipped
- TypeScript build and normal package build passed

After rebasing onto current `main` (`dbfd0b4`) and adapting test doubles to its new anti-slop lint fence:

- `npm run lint`
- `npm run test:types`
- 71/71 focused shutdown, signal, LLM, stdio MCP, HTTP daemon, and update-hook tests
- real hung HTTP daemon test verifies grace-expiry termination

No production QMD collection or index is used by these tests.


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #913, opened 2026-08-19
fix(search): bound expansion fanout and honor cancellation

rszrszrsz:
## Dependency

Draft: depends on #912. Until #912 merges, GitHub will also show its runtime commits in this PR. After it merges, this diff reduces to the search-policy and cancellation changes below.

## Summary

- bound query-expansion fanout once at the store boundary
- apply identical sanitization to fresh model output and legacy cached expansions
- exclude normalized restatements of the original query before consuming caps
- propagate caller and shutdown cancellation through expansion, embedding, vector fanout, and reranking
- make searches reject after abort instead of returning empty or partial success

## Expansion policy

The policy preserves source order and keeps:

- at most 6 variants total
- at most 2 variants per type (`lex`, `vec`, `hyde`)
- unique `(type, normalized text)` pairs

Normalization trims, collapses whitespace, and compares case-insensitively. Unknown types and empty text are rejected. The same policy runs after both cache reads and fresh generation, before any search fanout.

## Cancellation

Search observes the existing PR1 session abort signal plus an optional caller signal. Checks sit around the expensive boundaries:

- before and after expansion
- before and after embedding
- before every vector lookup and after the final lookup
- before and after reranking
- before returning RRF-only, empty, or partial results

`vectorSearchQuery` resolves the effective LLM, so stores without an attached LLM still observe shutdown from the default instance.

## Verification

Before rebasing onto the latest `main`:

- full Vitest: 1,255 passed, 78 skipped
- full Bun: 1,255 passed, 87 skipped
- normal build and CLI help/version passed

After stacking on the rebased #912 head:

- `npm run lint`
- `npm run test:types`
- 276 focused expansion-policy, cache, and cancellation tests passed; 13 existing tests skipped

No production QMD collection or index is used by these tests.


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #923, opened 2026-08-25
perf: avoid redundant CLI runtime spawn

ilepn:
# PR: perf: avoid redundant CLI runtime spawn

## Summary

- Import the compiled CLI in-process when the published package launcher is
  already running under the runtime selected by its lockfile.
- Preserve the existing spawned fallback when the launcher must switch between
  Node and Bun, and preserve the current source-checkout path.
- Align `process.argv[1]` with `dist/cli/qmd.js` before import so the CLI's
  existing main-module guard behaves as it does during a direct invocation.
- Add Node and Bun PID assertions proving that the compiled CLI stays in the
  launcher process.

The published `bin/qmd` launcher currently starts another copy of the selected
runtime even when npm or Bun has already invoked the launcher with that exact
runtime. Removing that duplicate process reduces the fixed cost of every QMD
CLI call without changing runtime selection or source-mode behavior.

## Windows benchmark

QMD 2.8.3, Windows 11, Node 26.7.0, Bun 1.4.0 and Bun 1.4.0 canary. Nine
interleaved keyword-search runs per runtime against the same four-document
synthetic corpus:

| Runtime | Existing launcher median | In-process launcher median | Change |
| --- | ---: | ---: | ---: |
| Node 26.7.0 | 223 ms | 167 ms | -25% |
| Bun 1.4.0 | 626 ms | 322 ms | -49% |
| Bun 1.4.0-canary.1 | 661 ms | 364 ms | -45% |

The Bun result also includes removal of a Windows-only `cmd.exe` hop, so these
percentages should not be generalized to POSIX hosts without separate data.
Model-backed cold starts remain dominated by native runtime/model loading; this
change targets the avoidable Windows launcher cost on every command.

## Verification

- `oxlint .`
- `tsc -p tsconfig.build.json --noEmit`
- focused Node/Vitest and Bun 1.4.0-canary.1 same-process PID assertions
- package-smoke coverage for both Node and Bun compiled-distribution wrappers
- the package smoke's POSIX executable-bit step is not natively runnable on
  Windows and remains a CI validation item
- `git diff --check`

The POSIX wrapper fixtures execute extensionless shebang files directly and
are not runnable through native Windows `execFileSync`; the new cross-platform
PID assertions pass under both runtimes. The package smoke now also exercises
both compiled-distribution wrapper paths.

## Local branch

- base: `dbfd0b4736aeaf761d1a16ca8e424f071df8feb9`
- commit: `dbc829a`
- branch: `perf/in-process-cli-launcher`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,11 @@
 
 - Added Oxlint lint fence.
 
+### Changed
+
+- Avoid a redundant runtime process on packaged CLI calls where the launcher is
+  already running under its selected Node or Bun runtime.
+
 ## [2.8.3] - 2026-08-16
 
 ### Security
--- a/bin/qmd
+++ b/bin/qmd
@@ -11,7 +11,7 @@
 import { spawn, spawnSync } from "node:child_process";
 import { existsSync, realpathSync } from "node:fs";
 import { dirname, resolve, sep } from "node:path";
-import { fileURLToPath } from "node:url";
+import { fileURLToPath, pathToFileURL } from "node:url";
 
 // Resolve symlinks so global installs (npm link / npm install -g) can find
 // the actual package directory instead of the global bin directory.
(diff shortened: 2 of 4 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #924, opened 2026-08-25
fix: harden daemon and embed-lock process identity on Windows

ilepn:
# PR: fix: harden daemon and embed-lock process identity on Windows

## Summary

- Replace bare daemon/embed PIDs with `qmd-process/v1` records containing the
  PID, role, and an OS-issued process-start token.
- Use Windows CIM process creation timestamps with hidden PowerShell execution,
  Linux `/proc/<pid>/stat` field 22, and `ps lstart` on other POSIX systems.
- Publish daemon identity only after the HTTP listener binds, observe early
  child exit, and verify the exact same identity immediately before signalling.
- Treat live pre-upgrade numeric state as unverifiable and fail closed instead
  of orphaning a daemon or stealing an active embed lock; role-specific POSIX
  command-line checks retain safe recognition where the owner can be proven.
- Pass the TSX loader as a file URL and hide spawned Windows process windows.
- Publish embed locks atomically, fail closed on fresh malformed state, and
  reclaim only an unchanged malformed lock after a 30-second grace period.

If the Windows CIM probe cannot obtain a process-start token, QMD writes a
legacy PID record and deliberately leaves it untouched while that PID remains
live. This degraded mode favors avoiding signals or lock theft against an
unrelated process; normal lifecycle management resumes after the process exits
or its ownership is independently reconciled.

## Why

The previous command-line inspection could not identify native Windows QMD
processes reliably. Bare PIDs are also unsafe after PID reuse and allowed
status/start/stop and embed-lock recovery to act on unrelated processes.

## Review-driven hardening

Two earlier adversarial Opus passes challenged upgrade compatibility, PID
reuse, protected-process probes, lock publication/recovery, runtime-specific
daemon launching, and Windows test coverage. A final Claude ultrareview then
found and verified two POSIX/signal-exit defects; both were corrected, and a
second ultrareview of the amended candidate completed with no findings. The
revision uses role-aware legacy checks, CIM creation timestamps, atomic lock
publication with a non-hardlink fallback, guarded malformed-state recovery,
and a Bun-native TypeScript child path.

## Verification

- `bun run lint`
- `bun run test:types`
- Node/Vitest and Bun 1.4.0-canary.1 process-identity and embed-lock tests:
  17 passed under each runtime
- native Windows `mcp http daemon` group: 13 passed under Node and 13 under
  Bun 1.4.0-canary.1
- `git diff --check`

## Local branch

- base: `dbfd0b4736aeaf761d1a16ca8e424f071df8feb9`
- commit: `355f4c8`
- branch: `fix/windows-process-identity`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,32 @@
 
 - Added Oxlint lint fence.
 
+### Changed
+
+- MCP daemon and embed-lock files now use structured process identity. A
+  pre-upgrade numeric file is acted on only when the live process command line
+  verifies the expected QMD role; otherwise it is left untouched.
+
+### Fixed
+
+- MCP daemon state and embed locks now bind a PID to its process start time
+  and role, preventing recycled or unrelated PIDs from being treated as their
+  owner. This also restores daemon and embedding-lock lifecycle checks on
+  native Windows.
+- Embed locks are published exclusively from a flushed temporary file. Fresh
+  malformed locks fail closed; unchanged malformed locks are reclaimed after
+  a 30-second grace period with a path-specific recovery message.
+- Legacy numeric state is recognized only when the live process command line
+  matches the expected QMD role, preserving safe upgrades on POSIX systems.
+- Source-checkout MCP daemons now pass the TSX loader as a file URL, allowing
+  detached daemon launches from Windows drive-letter paths.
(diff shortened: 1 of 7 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #928, opened 2026-08-26
fix(config): sanitize Windows absolute paths in index/config names

yotamleo:
## Impact

Windows users who point qmd at a collection or config using an absolute path (e.g. `D:\data\project`, or a UNC share) get a broken or invalid derived index/config filename -- the drive-letter colon and backslashes are written straight into the filename qmd computes for the on-disk config/DB, and Windows filenames cannot contain `:` or use `\` as anything but a path separator. This can produce a config file at an unintended nested path or fail outright depending on how the OS resolves the resulting string.

## Repro

At HEAD `dbfd0b4736aeaf761d1a16ca8e424f071df8feb9`, on Windows:

```
setConfigIndexName("D:\data\project")
```

`src/collections.ts:105` only special-cases names containing `/`; a Windows-style absolute path with no forward slash falls through to the `else` branch and is used verbatim as `currentIndexName`, so `getConfigFilePath()` (`src/collections.ts:127`) builds a path like `.../D:_data_project` is never sanitized -- it stays `D:\data\project.yml`, which contains a bare colon and backslashes. The same gap exists in `setIndexName()` in `src/cli/qmd.ts:220-228`, used when deriving the on-disk DB/index name from a CLI-supplied path.

## Fix

Detect a Windows-absolute path explicitly (`^[a-zA-Z]:[\/]` or a UNC `\server\share` prefix) in both `setConfigIndexName` (`src/collections.ts`) and `setIndexName` (`src/cli/qmd.ts`), and strip `:` and `\` alongside the existing `/` handling when deriving the sanitized filename. Added a regression test in `test/collections-config.test.ts` covering a Windows absolute path.

Found while maintaining a downstream fork.

### DIFF
--- a/src/cli/qmd.ts
+++ b/src/cli/qmd.ts
@@ -222,11 +222,16 @@ function mcpDaemonPaths(): { cacheDir: string; pidPath: string; logPath: string 
 
 function setIndexName(name: string | null): void {
   let normalizedName = name;
-  // Normalize relative paths to prevent malformed database paths
-  if (name && name.includes('/')) {
-    const absolutePath = pathResolve(process.cwd(), name);
-    // Replace path separators with underscores to create a valid filename
-    normalizedName = absolutePath.replace(/\//g, '_').replace(/^_/, '');
+  // Normalize relative paths to prevent malformed database paths. Windows
+  // absolute paths (C:\..., \\server\share) are already absolute -- skip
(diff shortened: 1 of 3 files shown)


PowderAddicts:
Dear tobi/qmd,

Thank you for your patience.

PR #906, opened 2026-08-17
fix(store): tokenize embeddings with the store model

0thernet:
## Problem

`generateEmbeddings()` and `maybeAdoptLegacyEmbeddingFingerprint()` select the store-local LLM for embeddings, but their chunking path calls the public `chunkDocumentByTokens()` helper, which initializes and tokenizes with the global default LLM. When a store selects another embedding model, chunk boundaries and legacy fingerprint verification can therefore use a tokenizer unrelated to the model that creates the vectors. The extra global lookup can also initialize an unrelated provider.

## Change

- add a private chunking helper that accepts the selected `LlamaCpp` instance
- use the store-selected LLM for embedding generation and legacy fingerprint adoption
- preserve the public `chunkDocumentByTokens()` API and its global-default behavior for standalone callers
- add regression coverage for both store paths, including a sentinel global tokenizer that must never run

The two regression tests fail on `facd35e` (v2.8.3) with `global tokenizer must not run` and pass with this patch.

## Verification

- TypeScript build typecheck
- Node: 1,145 passed, 78 skipped
- Bun: 1,145 passed, 87 skipped
- package smoke: compiled Node/Bun CLI and AST grammar resolution

PR #949, opened 2026-09-09
fix(store): tokenize embeddings with the store model

tobi:
Integrates and rebases #906 after review because the contributor branch cannot be updated with the current base.

Preserves the original commits and authorship. The fix routes embedding chunking and legacy fingerprint adoption through the store-local LLM, avoiding accidental initialization of the global default provider and keeping token boundaries aligned with the selected model.

Reviewed locally with:
- `bun run lint`
- `bun run test:types`
- `bun test --timeout 60000 --preload ./src/test-preload.ts test/store.test.ts test/sdk.test.ts` (341 passed)

Closes #906.


0thernet:
thanks @tobi !

Issue #952, opened 2026-09-14
vsearch takes ~70 s on an 11k-vector collection: exact scan runs per expanded query, expansions not deduplicated


jowitte:
## Summary

`qmd vsearch` takes ~70 s on a single ~11k-vector collection. Two things multiply:

1. **Collection-scoped searches use the exact-scan path** (`exactVecScanByHashSeq`, added for #791/#803) whenever the collection has ≤ `COLLECTION_VEC_EXACT_SCAN_MAX` (20,000) vectors. It computes `vec_distance_cosine(embedding, ?)` over `hash_seq IN (…)` lists, so sqlite-vec reads each vector individually from the chunk blobs via overflow pages.
2. **Expanded queries are not deduplicated.** `vectorSearchQuery` runs the exact scan once per expansion. For my query, the cached expansion had 31 vec/hyde entries, only 9 of them distinct, so 32 scans instead of 10.

## Environment

- qmd 2.8.3 (bun global), node-llama-cpp 3.20.0, Apple M5 Max, macOS 26.6.2
- Default collections: `akasha` (11,390 vectors, 5,787 docs) and `fumu-ch` (56 vectors)
- `vectors_vec` holds 52,877 rows in total (the rest belong to inactive/old documents); index.sqlite 290 MB

## Measurement

`qmd vsearch "Delegieren" -n 4`, no other qmd process running, expansion already cached. Timestamps per stderr/stdout line:

```
    0.1s Searching 32 vector queries...
    0.3s [node-llama-cpp] ggml_metal_library_init_from_source: error compiling source
   70.9s qmd://akasha/...   (first result)
```

So ~70 s is spent in the vector phase, ~2.2 s per query. Embedding the queries themselves takes a few ms each on Metal.

A `sample` of the process during that phase shows the main thread almost entirely in sqlite-vec column reads:

```
sqlite3_step
  sqlite3VdbeExec
    vec0Column
      vec0_get_vector_data
        blobReadWrite
          accessPayload
            getOverflowPage
              btreeGetPage → readDbPage → unixRead → pread
```

## Possible directions (just ideas)

- Deduplicate `queryTexts` in `vectorSearchQuery` (and probably in the hybrid path) before searching.
- Try capped ANN (`k` up to 4096) plus post-filter first, and fall back to the exact scan only when that yields fewer than `limit` results for the collection. That keeps the #791/#803 fix for small collections without paying the full scan for larger ones.
[...]

PR #962, opened 2026-09-17
Make search scale: bounded memory, fast default path

rikvanriel:
# Make search scale: bounded memory, fast default path

## Motivation

Searching a multi-thousand-file collection with large transcripts could
use gigabytes of heap (observed 4.2GB) and take over a minute per query
on CPU-only hardware. Three root causes: full document bodies were
loaded into JS for every candidate, every unbounded query used `all()`
which materialized full arrays in the V8 heap, and every search always
ran query expansion, vector search, embedding-model load, and LLM
reranking even when the keyword ranking was already decisive.
Reindexing had the same shape of problem: with no changes it still read
and hashed every file, costing minutes of IO on large collections.

## Results

On 5 sample queries over a ~13k-file collection, the slow queries went
from 54s/71s to 0.5-1.5s (35-60x faster), peak memory on the heaviest
query from 1.44GB to 164MB (~9x less), with 90%+ top-5 overlap against
the old full pipeline (4-5 of the top-5 files unchanged). An unchanged
collection now reindexes in seconds instead of minutes. A 10-query eval
of the optional centroid expansion changes the top-5 in 80% of queries
and pulls genuinely new documents (outside the old top-100) into the
top-5 on half of them.

## Patches

1. `feat(index): add mtime+size fast-path via file_sync_state` -
   reindex stats each file first and skips the read entirely when mtime
   and size match the cache, so an unchanged collection updates in
   seconds; files over 10MB and empty files are skipped.
2. `fix(search): bound hydration at DB level via substr, not JS` -
   caps each fetched body at 256KB in SQL as a first stop-gap against
   the heap blowup (40 winners max ~10MB instead of unbounded).
3. `fix(search,embed,index): stream large queries with iterate()` -
   converts every unbounded `all()` to streaming `iterate()`; on an
   index with 13k files and 89k chunk vectors this keeps tens of
   thousands of rows out of a single V8 array. Small bounded queries
   keep `all()`.
4. `feat(search): bound hydration and use chunk-level slices` -
   fetches only each winner's ranked 3.6KB chunk via
   `substr(doc, pos+1, len)`, so 40 winners stay under ~500KB total;
   chunk-only rerank approach copied from qmd-py's bounded hydration
   code and guard test.
5. `feat(search): score files from top non-overlapping chunks` -
   scores a file from its top-3 non-overlapping chunk hits (best vector
   distance to the query embedding first, adjacent overlapping chunks
   skipped so shared 15% overlap is not double-counted), approximating
   whole-file rerank at chunk cost.
6. `feat(cli, mcp, sdk): make fast search the default` - trusts the
   keyword ranking alone when the top FTS score clears 0.70, skipping
   expansion, vector search, model load, and rerank; full rerank stays
   available via `no-rerank false` / `rerank true`.
7. `feat(search): add centroid expansion` - optional Rocchio step that
   averages the top-3 winners' embeddings and searches the centroid for
   thematically related files the keywords missed, reusing stored
   vectors when available; auto-enables only when keyword recall looks
   thin (off by default otherwise).


rikvanriel:
This should fix #952 and related issues.

